PR #226, opened 2014-12-02
Issue 187 add proper equals method testing for config

### LINKED ISSUE #187: Missing tests for equality of config resources
We have no tests which test whether our implementations of the equality of resources like ```Accounts``` from e.g. ```config``` :package: are correct.

This is bad because other things like the (de-)serialization tests assume that this works.

f-porter:
Closes #187.
Sorry for the size of this PR but automated proper testing requires a few lines of code...


thechauffeur:
I reviewed the merge and the last commit -- both are simple (viewing the merge commit with ```git show``` which shows only the actual merged files rather than the github web interface). I will merge this now.

PR #231, opened 2014-12-10
[micro-optimization] Static StorageHTTP methods

aBothe:
See commit message.


jasinai:
> Is there any API doc that states StorageHTTP to be instantiated/non-static?

I don't think so.

Looks good. Can be merged.

cburkert:
Note that the API changes I made in #241 make it necessary to instantiate StorageHTTP, because it will contain connection state information.

aBothe:
Sure, already presumed that this pr affected mine :)

thechauffeur:
We decided that this PR will be closed in favor of #241. Review this instead.

Issue #232, opened 2014-12-11
Association: StorageServer <-> StorageVolume


jasinai:
According to [Storage Volume Doc](https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Configuration#storage-volume) there should be an association (the storage_server_id) to [Storage Server](https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Configuration#storage-server). But the [StorageServer Class](https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/StorageServer.java) doesn't have an ID.
Also the StorageServer**s** should have a method to find and get the StorageServer with the given ID.

thechauffeur:
Both Storage Server and Storage Volume do not need an id since both already have an unique identifier (URL and public_identifier). Maybe we will remove these later but for now we keep them because maybe we need them for Qabel Sync.

thechauffeur:
Storage Server and Storage Volume get linked with #241.

PR #233, opened 2014-12-12
 Try to fix wishes defined in #232

k-c13:
see #232 

Add function to create random id 	
Add id to constructor 	
Add find function 	
Add exception for already existing id 	
StorageServers::add: Check whether id already exists … 	
Update test 	
Create config class which creates random stuff 	
Implement new config class in SyncSettingItem class


thechauffeur:
Since #232 is closed now what else should this PR do or can this PR be closed? References to any issues would be nice.

src/main/java/de/qabel/core/config/SyncSettingItem.java
```diff
@@ -5,8 +5,8 @@
 /**
  * Class SyncSettingItem: Class to store common setting for synced item
  */
-public class SyncSettingItem {
-    private int id = 0;
+public class SyncSettingItem extends RandomValues {
```
cburkert: I don't think inheritance is the right way here. RandomValues is more a utility class.
Semantically, SyncSettingItem is definitely not a subclass of RandomValues.

src/main/java/de/qabel/core/config/RandomValues.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+		return result.toString();
+	}
+
+	/**
+	 * This method generates random integer
+	 * 
+	 * @return Random integer
+	 */
+	private int generateInt(int length) {
```
cburkert: What's the point of wrapping Random functions? I don't get it.

jan-schreib: The point is to build more possible errors. Nothing more, nothing less.

thechauffeur:
Since #232 is closed and I don't see any issue / discussion for the reason of this PR I will close it now. If someone thinks this is wrong: Open an issue, reference this PR and discuss it. We can reopen this PR if needed.

PR #239, opened 2014-12-28
Added a publicly accessible generation function for symmetric keys

### LINKED ISSUE #238: Implement easy-to-use symmetric key factory for storage
While implementing the storage client, I noticed that there is no painless way to create a new SecretKey for a caller of the new streaming crypto API.

I'll add a method to the key factory.

cburkert:
This is for example needed in the storage component which is going
to use the streaming crypto api.

Closes #238


jan-schreib:
Looks good. :+1:

PR #227, opened 2014-12-04
Enable all storage http tests

cburkert:
Depends on and MUST NOT be merged before Qabel/qabel-storage#14

After the storage server being patched, the ignore annotations in the storage http tests are no longer necessary.
Furthermore, this enables proper tests of unknown QSV handling by using unkown, yet syntactically valid UUIDs.

PR #241, opened 2015-01-05
Implemented storage API

### LINKED ISSUE #232: Association: StorageServer <-> StorageVolume

cburkert:
This implements the client API of the storage protocol.

EDIT: Closes #232


src/main/java/de/qabel/core/storage/StorageAction.java
```diff
@@ -0,0 +1,251 @@
...
+	public static StorageVolume createStorageVolume(StorageServer server) throws IOException {
+		StorageHTTP http = new StorageHTTP();
+		HTTPResult<StorageVolume> result = http.createNewStorageVolume(server);
+		StorageVolume volume = result.getData();
+
+		if (!result.isOk()) {
+			switch (result.getResponseCode()) {
+			case 503:
+				logger.info("Storage server reported capacity shortcoming.");
+				throw new RuntimeException("Storage server overloaded.");
```
L-Henke: Why do you throw a RuntimeException which might be uncatched instead of a IOException, which also fits the error and have to be catched by the caller?

cburkert: Good point. I'll change it.

src/main/java/de/qabel/core/storage/StorageBlob.java
```diff
@@ -1,4 +1,139 @@
...
+	 * @throws QblStorageBlobSizeExeeded if the blob exceeds the maximum size.
+	 * @throws QblStorageInvalidBlobName if name contains non-Base64url characters.
+	 */
+	public StorageBlob(InputStream bolbStream, String name)
+			throws IOException, QblStorageBlobSizeExeeded, QblStorageInvalidBlobName {
+		File temp = File.createTempFile("temp", ".bin");
+		temp.deleteOnExit(); // TODO is this safe?
+		FileOutputStream fos = new FileOutputStream(temp);
+		BufferedOutputStream bos = new BufferedOutputStream(fos);
+		byte[] copyBuffer = new byte[512];
```
L-Henke: Buffer size might be converted to a constant.

src/main/java/de/qabel/core/storage/StorageAction.java
```diff
@@ -0,0 +1,251 @@
...
+		if (!result.isOk()) {
+			switch (result.getResponseCode()) {
+			case 400:
+				logger.error("Volume probing failed because of syntactically invalid request.");
+				throw new RuntimeException("Unexpected response from storage server");
+			case 404:
+				logger.debug("Volume probing negative.");
+				return false;
+			default:
+				logger.error("Volume creation failed with unexpected response " + result.getResponseCode());
```
jasinai: It should be "Volume **probing** failed with unexpected response "

cburkert:
Note: Travis should pass if #227 merged.

src/main/java/de/qabel/core/storage/StorageAction.java
```diff
@@ -0,0 +1,251 @@
...
+	 * @throws InvalidKeyException
+	 * @throws QblStorageInvalidToken if write token is invalid.
+	 */
+	public static void uploadBlob(StorageVolume volume, StorageBlob blob, SecretKey key) throws IOException,
+			InvalidKeyException, QblStorageInvalidToken {
+		StorageHTTP http = new StorageHTTP();
+		CryptoUtils cryptoUtils = new CryptoUtils();
+
+		File encBlobFile = File.createTempFile("blob", ".enc");
+		FileOutputStream encBlobOutput = new FileOutputStream(encBlobFile);
```
aBothe: Why no MemoryStream? Is StorageHTTP providing no functionality to upload a file from a stream?

src/main/java/de/qabel/core/storage/StorageBlob.java
```diff
@@ -1,4 +1,140 @@
...
+
+import org.bouncycastle.util.Arrays;
+import org.bouncycastle.util.encoders.DecoderException;
+import org.bouncycastle.util.encoders.UrlBase64;
+
+import de.qabel.core.exceptions.QblStorageBlobSizeExeeded;
+import de.qabel.core.exceptions.QblStorageInvalidBlobName;
+
 public class StorageBlob {
+	public static final int MINIMUM_SIZE_BYTES = 1 * 1024 * 1024; // 1 MiByte
```
aBothe: Where did you take these values from?
Stepping further into the discussion, please pay attention to smaller files (e.g. Sync-Content) that shall be preserved as well - or should we pad the contents so it fits into the 1MB threshold?

cburkert: Well, these values are somehow arbitrary. Blob payloads can be small than the minimum threshold. Then they'll get padded. These threshold are actually padding ranges.
We decided to define a lower boundary to limit to hide meta data (file sizes).
The upper boundary is more about practical limitations but also a meta data obfuscation issue.

src/main/java/de/qabel/core/storage/StorageAction.java
cburkert: Well I'm now to Java Streaming stuff. So most certainly there are more efficient ways of doing this. Suggestions welcome.

cburkert:
If created a generic padding wrapper for InputStreams and used it to replace the naive padding mechanisms previously used for StorageBlobs.
The PaddedInputStream may also be used for drop message padding later.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -764,22 +766,35 @@ boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream,
...
 	 * @throws InvalidKeyException
 	 *             if key is invalid
+	 * @throws FileNotFoundException 
 	 */
-	boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream, SecretKey key, byte[] nonce)
-			throws InvalidKeyException {
+	public boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream, SecretKey key, byte[] nonce)
+			throws InvalidKeyException, FileNotFoundException {
+		FileInputStream fileInputStream = new FileInputStream(file);
+		return this.encryptStreamAuthenticatedSymmetric(fileInputStream, outputStream, key, nonce);
```
aBothe: fileInputStream should be inlined.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -831,12 +845,13 @@ boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream,
 	 * @throws InvalidKeyException
 	 *             if key is invalid
 	 */
-	File decryptFileAuthenticatedSymmetricAndValidateTag(InputStream inputStream, String pathName, SecretKey key)
+	public File decryptFileAuthenticatedSymmetricAndValidateTag(InputStream inputStream, String pathName, SecretKey key)
 			throws InvalidKeyException {
 		FileOutputStream fileOutput = null;
```
aBothe: Please avoid setting it null by default. Down there, if the FileOutputStream() ctor fails, make it null instead - or return, or create said temp file.

src/main/java/de/qabel/core/storage/PaddedInputStream.java
```diff
@@ -0,0 +1,112 @@
...
+	 * this method returns zero.
+	 * 
+	 * @return size of unpadded InputStream in bytes.
+	 */
+	public long getUnPaddedSize() {
+		long res = 0;
+		if (this.streamExhausted) {
+			res = this.totalReadBytes;
+		}
+		return res;
```
aBothe: Make this a one-liner, please.

src/main/java/de/qabel/core/storage/PaddedInputStream.java
```diff
@@ -0,0 +1,112 @@
...
+		if (this.paddingToWrite > 0) {
+			this.paddingToWrite--;
+			return FILL_BYTE;
+		}
+
+		// signal end of stream
+		return -1;
+	}
+
+	private void setPaddedSize() throws IOException {
```
aBothe: When is one allowed to set the pad size?
Also after having read a first raw stream byte?

src/main/java/de/qabel/core/storage/StorageAction.java
```diff
@@ -0,0 +1,251 @@
...
+	 * Creates a new storage volume on the given storage server.
+	 * 
+	 * @param server Server to host the volume.
+	 * @return New storage volume.
+	 * @throws IOException if server is overloaded.
+	 */
+	public static StorageVolume createStorageVolume(StorageServer server) throws IOException {
+		StorageHTTP http = new StorageHTTP();
+		HTTPResult<StorageVolume> result = http.createNewStorageVolume(server);
+		StorageVolume volume = result.getData();
```
aBothe: What about returning result.getData() directly instead of having an intermediary variable used for this?

src/main/java/de/qabel/core/storage/StorageAction.java
aBothe: Well, please consider using http://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html for this purpose. This might save a lot of IO time later on :P

src/main/java/de/qabel/core/storage/StorageAction.java
```diff
@@ -0,0 +1,251 @@
...
+	 * 
+	 * @param volume storage volume containing the blob.
+	 * @param blobName name of the blob.
+	 * @param key secret key to decrypt the downloaded blob.
+	 * @return file containing the decrypted blob.
+	 * @throws IOException
+	 * @throws InvalidKeyException
+	 * @throws QblStorageInvalidBlobName if given name contains non-Base64url characters.
+	 */
+	public static File retrieveBlob(StorageVolume volume, String blobName, SecretKey key) throws IOException,
```
aBothe: Imo, retrieveBlob should return a byte-Array, not a File path. I'm just thinking about running Qabel on Android gear where it could be likely that IO takes a long time joining all blobs to one raw file again.

cburkert: Blobs can be large, so holding them in memory as a whole is not practical, especially on mobile devices.

src/main/java/de/qabel/core/storage/PaddedInputStream.java
cburkert: The tertiary operator may indeed be more readable. I'll change it.

src/main/java/de/qabel/core/storage/StorageAction.java
aBothe: Okay, that rationalizes it :)

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -826,61 +840,66 @@ boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream,
...
+				}
+				fileOutput.write(encBytes);
+			}
+			try {
+				fileOutput.write(gcmCipher.doFinal());
+			} catch (IllegalBlockSizeException e) {
+				logger.debug("Decryption: File was encrypted with wrong block size.", e);
+				// truncate file to avoid leakage of incomplete or unauthenticated data
+				fileOutput.getChannel().truncate(0);
+				return false;
```
aBothe: Will fileOutput.close() be called here and some lines below?

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -2,97 +2,111 @@
...
+	 * @param token Token granting the right to upload blob to this storage volume.
+	 * @return OutputStream usable for uploading data.
+	 * @throws IOException
 	 */
-	public HTTPResult<?> upload(URL baseUrl, String publicIdentifier, String blobName, String token, byte[] blob) throws IOException {
-		URL url = addPathToURL(baseUrl, publicIdentifier + "/" + blobName);
-		HttpURLConnection connection = (HttpURLConnection) this.setupConnection(url);
-		connection.setRequestProperty("X-Qabel-Token", token);
+	public OutputStream prepareUpload(String publicIdentifier, String blobName, String token) throws IOException {
+		if (connection != null) {
```
aBothe: Could it be that a connection object may be reused and thus is not null but closed?

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -103,34 +117,41 @@
...
 	 * @param revokeToken The token, which is required to delete a blob or a qabel storage volume.
 	 * @throws IOException If something went wrong with the connection
 	 * @return HTTPResult  with empty data
 	 */
-	public HTTPResult<?> delete(URL baseUrl, String publicIdentifier, String blobName, String revokeToken) throws IOException {
-		URL url = addPathToURL(baseUrl, publicIdentifier + "/" + blobName);
-		HttpURLConnection connection = (HttpURLConnection) this.setupConnection(url);
-		HTTPResult<?> result = new HTTPResult<>();
+	public HTTPResult<?> delete(String publicIdentifier, String blobName, String revokeToken) throws IOException {
+		this.setupConnection(publicIdentifier, blobName);
```
aBothe: Why is there no check for whether the connection has been established already?

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -103,34 +117,41 @@
...
-	/**
-	 * Setups the connection to the given url.
-	 * @param url The whole url to the Qabel Storage Volume.
-	 * @return The open connection.
-	 * @throws IOException If something went wrong with the connection
-	 */
-	private URLConnection setupConnection(URL url) throws IOException {
-		return url.openConnection();
+	private void setupConnection(String publicIdentifier, String blobName) throws IOException {
+		StringBuilder resourcePath = new StringBuilder(publicIdentifier);
```
aBothe: What if a connection has been established already? Close the previous one? Throw something?

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -103,34 +117,41 @@
...
+		URL url = addPathToURL(server.getUrl(), resourcePath.toString());
+		connection = (HttpURLConnection)url.openConnection();
+	}
+	
+	private void setupConnection(String publicIdentifier) throws IOException {
+		this.setupConnection(publicIdentifier, null);
+	}
+	
+	private void closeConnection() {
+		connection.disconnect();
```
aBothe: Null-check missing!

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -103,34 +117,41 @@
...
-	 */
-	private URLConnection setupConnection(URL url) throws IOException {
-		return url.openConnection();
+	private void setupConnection(String publicIdentifier, String blobName) throws IOException {
+		StringBuilder resourcePath = new StringBuilder(publicIdentifier);
+		if (blobName != null) {
+			resourcePath.append("/");
+			resourcePath.append(blobName);
+		}
+		URL url = addPathToURL(server.getUrl(), resourcePath.toString());
```
aBothe: Is the addPathToURL method used multiple times? If not, you may inline this one.

src/main/java/de/qabel/core/http/StorageHTTP.java
```diff
@@ -157,14 +178,14 @@ private String parsePostResponse(InputStream inputStream) throws IOException{
 	 * @param response The parsed response.
```
aBothe: server param doc missing

src/main/java/de/qabel/core/storage/PaddedInputStream.java
```diff
@@ -0,0 +1,108 @@
+package de.qabel.core.storage;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class PaddedInputStream extends InputStream {
+	private static final byte FILL_BYTE = 0;
+	private InputStream stream;
+	private long totalReadBytes, paddingToWrite;
+	private long paddedSize, lowerBoundary, upperBoundary;
```
aBothe: Because these values are only set in the constructors, please make them `final`.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
aBothe: Okay, the finally-block and thus the fileOutput.close() will be executed even if the return happens in some try-block earlier, good to know! :D

src/main/java/de/qabel/core/http/StorageHTTP.java
cburkert: Every method calls closeConnection, so you cannot reuse a connection.

aBothe: Wait, each method only uses a connection once and only for exactly one call..then why share the connection object at all? - You even could make all these methods static, if possible!

src/main/java/de/qabel/core/storage/PaddedInputStream.java
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Creates an InputStream that pads the given InputStream to the given size. The stream is padded with the zero
+	 * byte.
+	 * 
+	 * @param stream Input stream to pad.
+	 * @param paddedSize Size of the padded stream in bytes.
+	 */
+	public PaddedInputStream(InputStream stream, long paddedSize) {
+		this(stream, paddedSize, paddedSize);
+		this.paddedSize = paddedSize;
```
aBothe: paddedSize should be validated via setPaddedSize() over here.

src/main/java/de/qabel/core/storage/PaddedInputStream.java
aBothe: Shouldn't this method be called validatePaddedSize() or so? Not sure whether a `set` method can't have any value passed via parameter.

aBothe: We've talked about it. I've RTFM now.

src/main/java/de/qabel/core/http/StorageHTTP.java
cburkert: Well, can't happen because connections are closed at the end of every call. But I'll put an assert into setupConnection to catch internal error.

src/main/java/de/qabel/core/http/StorageHTTP.java
cburkert: No, but it makes to code more readable to me.

aBothe:
Iirc, we yesterday couldn't answer on the question whether the StorageHTTP class should be thread-safe.
So, could anybody imagine any issue with leaving StorageHTTP thread-unsafe?

src/main/java/de/qabel/core/http/StorageHTTP.java
cburkert: Yes, but the server param is redundant anyway (its also a member variable) so I'll remove it from the signature.

aBothe: Even better :)

src/main/java/de/qabel/core/storage/PaddedInputStream.java
cburkert: What do you want to validate? If you have a look at setPaddedSize, you'll notice that it's not a simple setter but calculates the actual padded size in case of ranged padding.

aBothe: Sure, okay.

cburkert:
@f-porter Can you please have a look on the changes I made in the StorageVolume class? I removed the storage server id and put a reference to a StorageServer object in instead. I think we'd a type adapter for serialization then, right?

thechauffeur:
Also the tests and/or factories from your testing framework would need an update @f-porter.

cburkert:
Besides the test factory issue, all should be fine now.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -826,62 +840,67 @@ boolean encryptFileAuthenticatedSymmetric(File file, OutputStream outputStream,
 	 *            Pathname where the decrypted file will be stored
 	 * @param key
 	 *            Key which is used to en-/decrypt the file
-	 * @return The decrypted file or null if authentication tag validation
-	 *         failed or another error occured
+	 * @return true if successfully decrypted or true if authentication tag validation
```
jan-schreib: The second true should be a false, shouldn't it?

f-porter:
@cburkert: we will need a string "storageServerUrl" in the StorageVolume class for storing this value while the deserialization.
Concerning the TypeAdapter: If you add this "storageServerUrl" variable to StorageVolume class you can usually use the standard typeadapter, like it is the case right now, as the attributes of the class are standard types which are known by GSON. You only need a custom TypeAdapter if the de-/serialization shall differ from the standard behavior or if there are attributes of a non standard type.

cburkert:
@f-porter Ok, I'll do like you did it with contact and identity (with the protected constructor and so forth)

cburkert:
@f-porter Can you please review my last commit 32b2dee. I tried to adapt the concept used in the Contact/Identity relation, but the difference is that Contact has a type adapter and explicitly uses the constructor which is not setting the storageServer. So I'm not sure if this works...

f-porter:
It is possible to do this without an additional typeadapter but somehow that would be a workaround and could lead to complications later. Furthermore I will add a typeadapter for ```StorageVolume``` anyway in one of my next PRs as I need it to throw parsing exceptions. So it will be the best if you create a Typeadapter for that where you call your special constructor for deserialization.

cburkert:
@f-porter Please review my last changes. But I think this will do for now.

aBothe:
Any chance of having a dedicated method that puts together Blobs to a single byte[] or OutputStream in one dedicated method? 
I need such a method for Sync, and Sync is probably not going to be the only module which uses storage for storing data that won't exceed usual memory limitations.

thechauffeur:
@cburkert since @f-porter is not available this week you wanted @jan-schreib and @k-c13 to review specific things. Please update this PR.

jan-schreib:
If there are no more changes to come, I give this pullreq a :+1:

cburkert:
@thechauffeur I don't think we need to wait for @f-porter 's review. Only asked for his opinion because I used a slightly different approach than he used to in his type adapters, but mine works, too, and possible minor implementation style issues can be fixed afterwards.

thechauffeur:
@cburkert OK. @jan-schreib and I will review / merge this now.

aBothe:
Any results of reviewing this now?

Issue #68, opened 2014-10-06
Model Classes must be registered


Gottox:
To prevent code injections, prior to dynamical instatiations of Model Classes, Modules must register the Model Classes they use.

this may be done in the init method of a module.

k-c13:
@Gottox and I check the code and find out that in the moment nothing is todo. I will create a getter function for the class loader

Gottox:
We are using a dedicated ClassLoader for Classes. This classloader contains only classes from Module Jars. We assume, that we can trust qabel modules.

For an android port we need a refactoring at this point, but I think it's best if we delay this when we port it over.

jan-schreib:
This needs further discussion because #243 is so horribly broken.

thechauffeur:
Does #243 even has something to do with this issue?

thechauffeur:
According to  @Gottox this is obsolete but he wants to keep this open to further think about it.

thechauffeur:
@L-Henke do what you want with this. It is definitely not needed for the first release (as you can see from the label and the milestone).

PR #243, opened 2015-01-06
Create getter for class loader

k-c13:
Create getter for class loader
Remove static from class loader

see also #68


src/main/java/de/qabel/core/module/ModuleManager.java
```diff
@@ -98,7 +96,7 @@ public void startModule(Class<?> module) throws InstantiationException, IllegalA
 	
 	public void startModule(File jar, String className) throws InstantiationException, IllegalAccessException, ClassNotFoundException {
 		try {
-			ClassLoader cld = LOADER;
+			ClassLoader cld = classLoader;
 			cld.addURL(jar.toURI().toURL());
 			startModule(Class.forName(className, true, cld));
 		} catch (MalformedURLException e) {
```
aBothe: Does MalformedURLException also apply to startModule(), or just cld.addURL() ?

I'd shrink the try{}-Block then to the addURL-call, nothing else.

src/main/java/de/qabel/core/drop/DropDeserializer.java
```diff
@@ -26,7 +26,7 @@
         ModelObject m;
         Class<ModelObject> classModelObject;
         try {
-            ClassLoader loader = ModuleManager.LOADER;
+            ClassLoader loader = new ModuleManager().getClassLoader();
```
Gottox: ModuleManager must not be initialized multiple times. This will break things. We need a link between the ModuleManager and the DropController/DropDeserializer instead.

k-c13: Done

src/main/java/de/qabel/core/module/ModuleManager.java
k-c13: You are right

thechauffeur:
@jan-schreib and I reviewed, updated to current master and repaired history (and force pushed) this. However the DropDeserializer needs a DropController due to @k-c13 changes and we don't really know if this is a good idea. It brakes @cburkert code (abstract drop message format stuff) and maybe @f-porter would not like this, too. We have to look into this.

thechauffeur:
The tests all work and such, nevertheless @jan-schreib and I think this is broken by design since `ModuleManager` is now associated with `DropController` **and** the `DropDeserializer` now needs a `DropController` to get instantiated. This is weird. For example have a look at my [merge commit](https://github.com/Qabel/qabel-core/commit/675c7b32d00c82d1bc90a67f7014c65418d7a566) where `AbstractBinaryDropMessage` now needs a `DropController`, too (because it needs a `DropDeserializer`).

DO **NOT** MERGE THIS!

thechauffeur:
Also `DropController.retrieve()` ever since creates a new `DropController` which now always creates a new `ModuleManager`.

jan-schreib:
> ModuleManager must not be initialized multiple times.

This pullrequest moves the mistake that was pointed out by @Gottox deeper into the code.
Now even a new DropController creates a new ModuleManager. This code is broken.

thechauffeur:
Closed because a complete redesign would be needed.

PR #245, opened 2015-01-12
242 add versionheader to drop message

### LINKED ISSUE #242: Implementation not compliant with Drop Message transport format spec
The [spec](https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Drop#transport-format) demands a version header which is currently missing in the implementation.

jasinai:
Concatenates the header with the encrypted message and removes it from the received message.
Fixes #242

### DIFF
--- a/src/main/java/de/qabel/core/drop/DropController.java
+++ b/src/main/java/de/qabel/core/drop/DropController.java
@@ -300,7 +300,7 @@ public class DropController {
 	}
 
 	/**
-	 * Concatenates the header byte array to the message byte array.
+	 * Concatenates the header to the message
 	 * @param header
 	 * @param message
 	 * @return The message with the prepended header
--- a/src/main/java/de/qabel/core/drop/DropController.java
+++ b/src/main/java/de/qabel/core/drop/DropController.java
@@ -191,8 +191,9 @@ public class DropController {
 			cryptedMessage = encryptDrop(serialize(message),
 					contact.getEncryptionPublicKey(),
 					contact.getContactOwner().getPrimaryKeyPair().getSignKeyPairs().get(0));
+			byte[] cryptedMessageWithHeader = concatHeaderAndEncryptedMessage((byte) message.getVersion(), cryptedMessage);
 			for (DropURL u : contact.getDropUrls()) {
-				result.addErrorCode(http.send(u.getUrl(), concatHeaderAndEncryptedMessage((byte) message.getVersion(), cryptedMessage)));
+				result.addErrorCode(http.send(u.getUrl(), cryptedMessageWithHeader));
 			}
 		} catch (InvalidKeyException e) {
 			logger.error("Invalid key in contact. Cannot send message!");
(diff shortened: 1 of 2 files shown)


aBothe:
Looks fine to me :)

jan-schreib:
What about a tests for your new methods? Trivial, but you know the drill. :)

jasinai:
> What about a tests for your new methods? Trivial, but you know the drill. :)

Yes, perhaps it's a good idea

src/main/java/de/qabel/core/drop/DropController.java
```diff
@@ -293,4 +298,30 @@ private String decryptDrop(byte[] cipher, QblPrimaryKeyPair keypair, QblSignPubl
 		CryptoUtils cu = new CryptoUtils();
 		return cu.decryptHybridAndValidateSignature(cipher, keypair, signkey);
 	}
+
+	/**
+	 * Concatenates the header byte array to the message byte array.
```
jan-schreib: According to the signature, the header is not an array.

src/main/java/de/qabel/core/drop/DropController.java
```diff
@@ -188,8 +191,9 @@ public DropResult send(DropMessage<? extends ModelObject> message, Collection<Co
 			cryptedMessage = encryptDrop(serialize(message),
 					contact.getEncryptionPublicKey(),
 					contact.getContactOwner().getPrimaryKeyPair().getSignKeyPairs().get(0));
+			byte[] cryptedMessageWithHeader = concatHeaderAndEncryptedMessage((byte) message.getVersion(), cryptedMessage);
```
cburkert: The version returned by DropMessage's getVersion is not version required here.
We need to version of the **binary** message format here (see [doc](https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Drop#header)).
Please hard-code the version to 0 until we implement proper version multiplexing.

jasinai:
@jan-schreib: do you really want these tests? I would test java internals. My tests look like this: https://gist.github.com/jasinai/f1ce89e74c640ab7a43f, so if you really want them, I have them.

thechauffeur:
Later -- when we have different versions -- we will need a check with different handling according to the version and we will need tests for that.
For now consider writing one single test which creates a drop message, adds the header using your new method, removes the header using your other new method and then checks whether the drop message is the same message prior to adding the header. I would leave out the tests you mentioned above @jasinai.

After adding such a test I :+1: for the merge.

jan-schreib:
@jasinai Sorry for the missunderstanding. Of course I don't want you to test Java internals. I don't want to make your work harder. :)
@thechauffeur is right about the test method and I :+1 for the merge also when finished.

jasinai:
The test could be a little bit smaller, if serialize and deserialize from DropController were not private. :(

jan-schreib:
Ok for now. @jasinai please create a ticket for the code duplication in the tests. It's not critical right now, but has to be done in the future.

PR #256, opened 2015-01-21
Issue #255 Check core of deprecated code 

### LINKED ISSUE #255: Check core of deprecated code
Check existing code of deprecated code und replace it

k-c13:
Close #255

### DIFF
--- a/src/main/java/de/qabel/core/crypto/CryptoUtils.java
+++ b/src/main/java/de/qabel/core/crypto/CryptoUtils.java
@@ -554,13 +554,14 @@ public class CryptoUtils {
 		}
 
 		// Decrypt RSA encrypted AES key and decrypt encrypted data with AES key
-		byte[] rawAesKey = rsaDecrypt(encryptedAesKey,
-				privKey.getQblEncPrivateKeys().get(0));
-		if (rawAesKey != null) {
-			logger.debug("Message is OK!");
-			return new String(decryptSymmetric(aesCipherText, 
-					new SecretKeySpec(rawAesKey, SYMM_KEY_ALGORITHM)));
-		}
+		for (RSAPrivateKey key : privKey.getQblEncPrivateKeys()) {
+            byte[] rawAesKey = rsaDecrypt(encryptedAesKey, key);
+            if (rawAesKey != null) {
+                logger.debug("Message is OK!");
+                return new String(decryptSymmetric(aesCipherText,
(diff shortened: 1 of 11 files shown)


aBothe:
Imho, it's okay to use the `.get(0)` in the tests.
What I'm not sure about is to keep using the first key only in the core code.

k-c13:
> What I'm not sure about is to keep using the first key only in the
> core code.
When you check the deprecated method you will see that the always use
the first key. 
If not we have to change the code in the methods which use it

aBothe:
Sure, well okay, these can be changed later on then. So :+1: for merge.

jan-schreib:
@L-Henke wrote in his [commit](https://github.com/Qabel/qabel-core/commit/4964cc3e8401dfcb963628953e143d6d373acb10) (which added the `@deprecated`) that this needs to be done. Code looks fine to me. Ok for merge.

L-Henke:
Changing the calls to ```.get(0)``` is not enough to adapt to the changes, because the "new" methods can return a list of keys and thus (at least sometimes) the whole list has to be iterated and not just the first element.

L-Henke:
I think this PR should also remove the deprecated methods, since it removes all of its usages.

k-c13:
Ok. I will do it.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -554,13 +554,14 @@ public String decryptHybridAndValidateSignature(
...
 
 		// Decrypt RSA encrypted AES key and decrypt encrypted data with AES key
-		byte[] rawAesKey = rsaDecrypt(encryptedAesKey,
-				privKey.getQblEncPrivateKey());
-		if (rawAesKey != null) {
-			logger.debug("Message is OK!");
-			return new String(decryptSymmetric(aesCipherText, 
-					new SecretKeySpec(rawAesKey, SYMM_KEY_ALGORITHM)));
-		}
+		for (RSAPrivateKey key : privKey.getQblEncPrivateKeys()) {
```
L-Henke: The indentation is a little bit weird, because the new ```for``` line uses tabs while the old code uses spaces.

L-Henke:
Technically this PR seems correct, so :+1: for that.

jan-schreib:
I'm fine with it, too. :+1:

PR #257, opened 2015-01-23
Fixed inconsistency in QblPrimaryPublicKey

cburkert:
The two lists for signing and encryption sub-public keys are not updated by the QblPrimaryKeyPair constructor.

### DIFF
--- a/src/main/java/de/qabel/core/crypto/QblPrimaryKeyPair.java
+++ b/src/main/java/de/qabel/core/crypto/QblPrimaryKeyPair.java
@@ -1,5 +1,6 @@
 package de.qabel.core.crypto;
 
+import java.security.InvalidKeyException;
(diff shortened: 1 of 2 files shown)


cburkert:
Travis failed, but it seems unrelated to my PR, because it says `npm install` had failed. ??

jan-schreib:
Travis is happy and I give this a :+1:

L-Henke:
Looks like I forgot to add the keys to the QblPrimaryPublicKey. :+1: from me, even if I don't know if the test is really necessary.

cburkert:
@L-Henke The test exposes the bug. Isn't it common practice to add such tests?

L-Henke:
I know, but I'm not sure if we ever re-introduce that bug, so we might not need the test. But of course having more tests doesn't harm...

PR #260, opened 2015-01-26
Remove checkout of a specific qabel-storage branch

jasinai:
Removes the checkout of the protocol_update branch, because the master branch is up-to-date.


jan-schreib:
We talked about this in the meeting. No big review needed.

PR #258, opened 2015-01-23
Support for multiple drop message binary transport formats

cburkert:
This implements
* a basic binary drop message handling API
* handling of binary drop message in the current version (0)

EDIT:
Implements message padding and closes #251 
Also implements a transport message handling thus closes #61 

Depends on Qabel/qabel-drop#14 to update the maximum drop message size on the drop server.

PR #261, opened 2015-01-26
Hotfix: Increased size of too-big-drop-message

cburkert:
Solves current test failure of test postMessageTooBig.

Cherry picked from #258 to be faster reviewable ;-)
Not needed if #258 got merged first.


k-c13:
Looks good. Thanks for that

PR #262, opened 2015-01-26
Class diagram

k-c13:
Update class diagram. Also the image in Qabel-Doc


jan-schreib:
I think you made a branch from the `k-c13:deprecated-code` branch. Eight of nine commits are not needed here.

k-c13:
Sorry for that. I do not know how that happened. I will do it new

PR #264, opened 2015-01-27
Update gradle for integration test

k-c13:
Here is the updated gradle configuration

The only problem is the naming of the nodejs command. On my Debian machine it is nodejs.


jan-schreib:
The nodejs command is not the problem right now. Read the travis output.
It says:
`java.io.IOException: Cannot run program "./drop_server.py" (in directory "../qabel-drop"): error=2, No such file or directory`

k-c13:
The problem is that it need the complete Qabel repository with the submodules drop and storage to work with. The test on Travis have to change but I do not have rights to do it.

But local it works inside the qabel repository

jan-schreib:
Closed in favor of #274

PR #268, opened 2015-01-31
Let ModuleManager.settings be a single Settings-instance, not a Set of Settings.

### LINKED ISSUE #266: Why is ModuleManager.Settings a set of settings, not just a single 'Settings'-instance?
[tl;dr](https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/module/ModuleManager.java#L54)

A Settings-object already contains local&synced settings for all modules,
so why are we making up a further dimension to store multiple sets of module settings?

aBothe:
Fixes #266.

### DIFF
--- a/src/main/java/de/qabel/core/module/ModuleManager.java
+++ b/src/main/java/de/qabel/core/module/ModuleManager.java
@@ -44,18 +44,11 @@ public class ModuleManager {
 		return this.storageConnection;
 	}
 
-	/**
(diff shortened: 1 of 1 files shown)


L-Henke:
Since @Gottox confirmed in #266 that no ```Set``` of settings is required: :+1:

jan-schreib:
:+1:

Issue #265, opened 2015-01-29
Why doesn't Contact inherit from SyncSettingItem?


cburkert:
While refactoring the drop API, I needed a common class for Contact and Identity to use it as sender in a DropMessage. So I considered introducing an abstract class which both Contact and Identity could be derived from.
But, I noticed that Identity inherits from SyncSettingItem but Contact does not.
Is this right? Should contacts be synchronized, too?

aBothe:
Indeed, seems to be an issue.

f-porter:
you are right. If we want to sync the contacts the same way like the other synced items then it should also inherit from SyncSettingItem. Sure, syncing it the same way would make sense IMO.
But that's not described in the documentation (https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Configuration#settings) where contacts are also missing in the JSON specified for SyncedSettings.

k-c13:
The reason for that was that contact not really a part of the configuration. So no one add this to that class. 

I update document https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Contact-Drop-Messages

cburkert:
@k-c13 If contact is not a setting, which I understand, then what do you think about changing the SyncSettingItem to a SyncItem that is semantically not limited to settings or - if we need a Sync**Setting**Item introduce an additional, more general SyncItem?

thechauffeur:
We have two PRs fixing this issue. @cburkert your PR is not linked here. Please discuss which one should be used. Thanks.

cburkert:
My PR #269 introduces a common base class for Contact and Identity which itself inherits from SyncSettingItem, therefore Contact is a SyncSettingItem, too.

k-c13:
> @k-c13 If contact is not a setting, which I understand, then what do
> you think about changing the SyncSettingItem to a SyncItem that is
> semantically not limited to settings or - if we need a SyncSettingItem
> introduce an additional, more general SyncItem?
You are right. You shall change it in your PR

cburkert:
@k-c13 Unfortunately, it is not that easy because Contact and Identity must have the same base class, so I cannot make Contact a SyncItem and Identity a SyncSettingItem. Maybe we can realize the "setting"-property through an interface, but it is too big a change to squash it into my PR #269

thechauffeur:
We do not distinguish between `contact` or `setting`. Technically they are the same for us. #269 does reflect this.

thechauffeur:
[...]

PR #267, opened 2015-01-30
Issue #265  Why doesn't Contact inherit from SyncSettingItem?

k-c13:
See #265 

Add extended SyncSettingItem class 	
Update TypeAdapter for using extended SyncSettingItem class 	
Update test

PR #269, opened 2015-02-01
Refactored Drop API

### LINKED ISSUE #247: API for sending a drop message / signing of drop message
I browsed through the test code of the drop component found some things
* first of all I find it really complicated to send a drop message:
I have to instantiate a drop message by my own and set several parameters (of which some don't depend on the senders state, e.g. the timestamp, version, acknowledgeID) before I can finally send it.
Is there a reason why that's not encapsulated somewhere (e.g. DropController)? Or is it just because of the last milestone's urgency?
* Additionally I wonder why the encrypted drop message gets signed with the sign key of the recipient instead of the sender (in class DropController:
```cryptedMessage = encryptDrop(serialize(message),
					contact.getEncryptionPublicKey(),
					contact.getContactOwner().getPrimaryKeyPair().getSignKeyPairs().get(0));```)
Or do I miss something here?
* The same happens when decrypting the message, the private key of a contact is used where the one of an identity should be used.

Edit:
Wait, now I see that the contact owner is used. so forget the last two points ;)

### PR DESCRIPTION
**Pull request by cburkert**
I refactored drop API with particular focus on
* use model classes instead of serialization representations wherever possible
* make DropMessage construction easier

This PR
* introduces a abstraction layer called Entity above Contact and Identity that contains common properties like drops, public key getters
* introduces an general mapper for Entity lookups by key identifier and uses it for Contacts and Identites
* implements acknowledge ID generation using UUIDv4
* implements detection of spoofed sender information in DropMessages
* complains about unexpected drop message version by throwing parser exception instead of treating the version field like a meaningless field

Closes #247 

EDIT: Added version checking "functionality" description

### CODE REVIEW

File: src/main/java/de/qabel/core/config/Contact.java [hunk 1]
```diff
@@ -170,18 +87,19 @@ public Contact(Identity owner)
 	 * Attention: This constructor is intended for deserialization purposes. The contactOwner needs to be set afterwards
 	 * @param ownerKeyId
```
aBothe: Thanks for the 'Attention' rationale in the comment! :)
The `@param`s for the other two parameters are missing.

src/main/java/de/qabel/core/config/EntityMap.java
```diff
@@ -0,0 +1,68 @@
+package de.qabel.core.config;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+abstract public class EntityMap<T extends Entity> {
```
aBothe: Despite I know that a map is some sort of lookup, it doesn't get clear what entities are mapped to (their keyIdentifiers?). Please mention this in the comment.

src/main/java/de/qabel/core/drop/DropDeserializer.java
```diff
@@ -18,13 +18,15 @@
                                                 JsonDeserializationContext context) throws JsonParseException {
 
         int version          = json.getAsJsonObject().get("version")       .getAsInt();
+        if (version != DropMessage.getVersion()) {
+        	throw new JsonParseException("Unexpected version: " + version);
```
aBothe: This new functionality should be mentioned in the PR description as well. Not sure whether the logic should at least try to examine readable info -- perhaps some versions are backward compatible or so. But well, not in this PR! :)

src/main/java/de/qabel/core/drop/DropMessage.java
```diff
@@ -1,80 +1,102 @@
...
+    		return false;
+    	}
+    	this.sender = sender;
+    	return true;
     }
 
-    public void setModelObject(Class<T> modelObject) {
-        this.modelObject = modelObject;
+    public Class<T> getModelObject() {
+        return (Class<T>)data.getClass();
```
aBothe: This seems way more elegant than explicitly storing some class name in an extra property.
Despite that, what if data happens to be `null`?

In my first Sync-approach, I'm handing in a `new DropMessage<SyncDropMessage>()` without changing anything else into DropController.sendAndForget() - so there will be no data set. I'm not sure whether this is sufficient though, but using DropMessage that way seems reasonable to me.

aBothe:
In spite of the initial things I've critizied, the rest of this pr looks fine to me. Now let's place bets on how long it'll take for it to get merged.

thechauffeur:
So this also closes #265 and also implements acknowledge ID generation and also implements spoofing detection.
Well... it would be good if you split your pull requests more often or write issues like "Implement spoofing detection" and assign them to yourself and wait with implementing them until merging (fixing) your current PR (if they depend on each other).
Your PRs are not really nice to review and also implement / fix stuff without a corresponding issue. Have we even decided on the acknowledge IDs yet? Also referencing all effected issues would be really really nice...

src/main/java/de/qabel/core/config/Contact.java
cburkert: Well, I didn't write the attention note. But you're right, the other params are missing. I'll remove the one param comment because the are self-explaining and redundant anyway.

src/main/java/de/qabel/core/drop/DropDeserializer.java
cburkert: Well, I don't see much of a new functionality in this, but I'll add it to the description.

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: I don't see why you should pass null as payload instead of an instance of SyncDropMessage.

cburkert:
@aBothe I cannot change the send API to Entity as recipient type that easy because, the Contact-specific contactOwner property is used to get the private keys for signing. I could use the sender reference in the DropMessage instead but have to think about the implications first.
I'd rather like to work this out in another PR, but I understand your thoughts.

cburkert:
@thechauffeur I'm not sure if this PR closes #265 because this is not a feature request issue but a discussion issue and there are points in that discussion (e.g. about whether or not a Contact is a setting) which are not magically solved even if PR is eventually merged.

The acknowlege IDs are [documented in the spec](https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Drop#acknowledging). Since I had to return any acknowledge ID in the getter to avoid syntactical errors, I thought a one-liner that creates UUIDs is at least as good a solution as returning an empty string.

The spoofing detection is also a really insignificant by-product. Taking four LOCs it should not complicate this PR much.

I'm sorry that this PR is large, but since it is a refactoring PR, I couldn't split it up while maintaining any meaning.

thechauffeur:
@cburkert OK, I see.

cburkert:
Changed the serialization of the properties Contact inherits from SyncSettingItem in 60a97e9366373772efbd00368eaa7a4518ea2414 to the flat/unnested approach in #267 
See commit message.

Now I'll make this "mergeable" again.

cburkert:
By the way, this PR is ready for merging again.

jan-schreib:
CALM DOWN WE ARE LOOKING AT THIS!!1!

thechauffeur:
@jan-schreib yeah well... did we? I mean we did but somehow we did not respond here -- why? In any case this should be reviewed and then finally merged :D

thechauffeur:
> By the way, this PR is ready for merging again.

Well, not anymore :P

jan-schreib:
> Well, not anymore :P

@cburkert U mad, bro?

src/test/java/de/qabel/core/config/ContactTestFactory.java
```diff
@@ -20,8 +20,9 @@
...
 	@Override
 	public Contact create() {
-		Contact contact = new Contact(idFactory.create());
-		contact.setPrimaryPublicKey(qppkFactory.create());
-		return contact;
+		return new Contact(
+				idFactory.create(),
+				null,
+				qppkFactory.create());
 	}
```
thechauffeur: No need to wrap lines here, you know.

src/test/java/de/qabel/core/drop/DropControllerTest.java
```diff
@@ -16,142 +16,99 @@
...
-
-        recipientContacts = new Contacts();
-        recipientContacts.add(recipientContact);
-
-        dropController = new DropController();
-		
-	}
+    
+    @Before
+    public void setup() throws MalformedURLException, QblDropInvalidURL, InvalidKeyException {
```
thechauffeur: These two lines have not really changed. Please do not just change the whitespaces in lines you don't touch.

src/main/java/de/qabel/core/drop/DropMessage.java
```diff
@@ -1,80 +1,102 @@
...
     }
 
-    public void setData(T data) {
-        this.data = data;
+    /**
+     * Enable/disable acknowledge request for this drop message.
+     *
+     * @param enabled
+     */
+    public void enableAcknowledgeing(boolean enabled) {
```
thechauffeur: Since the default is `enabled` this name is misleading.

Also it does not check whether the caller already had enabled acknowledging -- and this already generated an acknowledge id -- by using the normal constructor almost everyone would use (see above).

src/main/java/de/qabel/core/drop/DropMessage.java
```diff
@@ -1,80 +1,102 @@
...
-        setData(data);
-    }
-
-    public int getVersion() {
-        return version;
-    }
-
-    public void setVersion(int version) {
-        this.version = version;
+    public DropMessage(Entity sender, T data) {
```
thechauffeur: This magically enables acknowledging which is not even documented. I'm also not sure if this should be the default. Also see comment [below](https://github.com/Qabel/qabel-core/pull/269/files#r24431178).

Maybe we should change this constructor and add a third needed parameter `boolean enableAckowledging`.

thechauffeur:
Besides my comments @jan-schreib and I ACK this PR. Also a new merge with master would be nice.

jan-schreib:
I think its more sane to disable acknowledgement by default. Maybe add a new constructor with a 3rd parameter, but I don't like that idea much. Needs further discussion.

src/main/java/de/qabel/core/drop/DropMessage.java
jan-schreib: I think its more sane to disable acknowledgement by default. Maybe add a new constructor with a 3rd parameter, but I don't like that idea much. Needs further discussion.

src/test/java/de/qabel/core/drop/DropControllerTest.java
cburkert: Oh, that must be a bad relict of my conflict resolution. I guess you don't want me to change it back, do you?

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: Ok, I'll make no-acknowledging the new default. I'd say, in `send` one should then call

    dropMessage.enableAcknowledgeing(true);

and off it goes.
No need for changing constructors then.

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: I'll prevent re-generation of ackIds if they are already set. But if NOACK will be be the new default, this function should no longer be misleading, so I can keep the rest as it is.

cburkert:
PR is ready for merging again.

src/test/java/de/qabel/core/drop/DropControllerTest.java
thechauffeur: No, but just because you asked so nice :P

src/main/java/de/qabel/core/drop/DropMessage.java
thechauffeur: Aye.

src/main/java/de/qabel/core/drop/DropMessage.java
thechauffeur: Aye.

src/main/java/de/qabel/core/drop/DropMessage.java
```diff
@@ -1,80 +1,107 @@
...
-        this.data = data;
+    /**
+     * Enable/disable acknowledge request for this drop message.
+     * Acknowledging is disabled by default.
+     *
+     * @param enabled true enables acknowledging
+     */
+    public void enableAcknowledgeing(boolean enabled) {
+    	if (enabled && this.acknowledgeId != NOACK) {
+    		this.acknowledgeId = generateAcknowledgeId();
```
thechauffeur: This will overwrite an existing acknowledgeId **and** will render it useless because acknowledging can never be enabled once NOACK has been set.

src/main/java/de/qabel/core/drop/DropMessage.java
thechauffeur: BTW, do you plan to implement this in this PR? If so we could use a (some) test(s) `DropControllerTest` for `send` and test the new acknowledgement true/false behaviour -- which could later be extended to test acknowledging `DropMessages`.

thechauffeur:
@cburkert if you do a `git show 7fb8fb7bb7fa76c3f4d8154a17c9c15c62b4a06e` you can see -- again -- a lot of white space only changes in the diff. I don't know how you are able to introduce these even in a simple merge commit. Maybe you should use things like the git console or git gui tools like `gitk` more often instead of relying on your IDE :P

I would merge this PR anyway (if not for the bug mentioned above), nevertheless this is somewhat annoying (of course it is not really your fault though -- it is the mix of tabs and spaces).

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: Oh, stupid me. I'll fix it and write a test.

thechauffeur: :+1:

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: No, I didn't plan to implemented acknowledged sending in this PR. I'm putting the acknowledge enabling tests into a dedicated test without dependencies to DropController.

cburkert:
@thechauffeur Believe me, I don't use my IDE for git stuff at all. The reason for those trailing white spaces being removed is as simple as stupid: I rewrote almost the entire Identity class during the introduction of the super class and discarded some trailing white spaces around my modifications. Conflicting changes by @f-porter in the same file caused git automatic conflict resolution to take my stripped version.
Long story short: I didn't pick those changes manually for the merge commit but after auto-resolving it really looks stupid, without any doubt. I promise to take care of this in the future...

src/main/java/de/qabel/core/drop/DropMessage.java
cburkert: Fixed.

jan-schreib:
I don't care about the whitespaces right now. You fixed the last bug and this is good to go. Merge.

PR #272, opened 2015-02-02
Replace DropController by DropActor, an implementation of the Ackack EventActor.

Gottox:
Open ToDos:

* [x] Tests
* [x] Updating Javadoc see https://github.com/Qabel/qabel-core/pull/272/files#r24000488
* [x] Adapt changes to other Repositories (qabel-desktop, qabel-helloworld-module)
* [x] Update Javadoc see https://github.com/Qabel/qabel-core/pull/272/files#r25595552
* [x] Fix Testrun see https://travis-ci.org/Qabel/qabel-core/builds/53625534
* [x] Fix ackack for Travis Test

PR #273, opened 2015-02-02
Make DropCallback public

aBothe:
Make DropCallback public because it must be passed via the public DropController.register() method.

### DIFF
--- a/src/main/java/de/qabel/core/drop/DropCallback.java
+++ b/src/main/java/de/qabel/core/drop/DropCallback.java
@@ -1,5 +1,5 @@
 package de.qabel.core.drop;
 
-interface DropCallback<T extends ModelObject> {
+public interface DropCallback<T extends ModelObject> {
 	void onDropMessage(DropMessage<T> message);
 }


jan-schreib:
I don't get it. Why is that needed? I don't see anything that needs it.

aBothe:
Not sure if this is the best solution, but when setting up Sync, I'm registering my custom DropMessage-class in DropController. The DropListenerTests do this as well, just saying.

```
// Register for sync notifications
modMgr.getDropController().register(SyncDropMessage.class, new DropCallback<SyncDropMessage>() {
	@Override
	public void onDropMessage(DropMessage<SyncDropMessage> message) {
		enqueueSync();
	}
});
```

aBothe:
I mean otherwise, if DropCallback shall be kept private, DropController.register could only take DropQueueCallbacks (from e.g. external code) - which should be made more obvious then in some way.

jan-schreib:
I don't understand that bunch of code where DropCallback is used and I don't understand this whole clusterfuck of generics and stuff any more.
@Gottox maybe you have something to say to this pullrequest.

aBothe:
that `MyDropCallback.run` method is called as soon as said `MyDropMessage` arrives at your Drop. Fairly comprehensible. To let DropController know that there is such MyDropMessage and MyDropCallback.run, you've gotta call DropController.register().
Problem: DropController.register() is public, the abstract DropCallback interface isn't. DropQueueCallback is the only implementation of that private DropCallback interface.
Thus, DropController.register() can practically only take DropQueueCallbacks atm. This isn't mentioned anywhere.

Gottox:
DropCallback will go away anyway :)

jan-schreib:
@aBothe if the answer from @Gottox is ok for now, please close this pullrequest.
After #272 is merged you will have a new way to implement your code, I think.

aBothe:
Okay then

PR #274, opened 2015-02-03
Updated gradle and travis config.

jan-schreib:
Moved the build directories.
Gradle starts the servers now, no need for doing it by hand anymore.

Based on @k-c13 work.


thechauffeur:
:+1:

PR #277, opened 2015-02-04
Changed size of drop message size test

### LINKED ISSUE #276: Change size in drop message size test to a meaningful value
See https://github.com/Qabel/qabel-core/blob/master/src/test/java/de/qabel/core/http/DropHTTPTest.java#L84 versus https://github.com/Qabel/qabel-drop/blob/master/drop_server.py#L55

thechauffeur:
to one byte greater than the server accepts.

Closes #276


jan-schreib:
Trivial change, :+1: and merge.

PR #279, opened 2015-02-05
Javadoc cleanup

jan-schreib:
I wanted to create a javadoc with `./gradlew javadoc` and noticed a couple of warnings.
I fixed the warnings and found some missing things.
Maybe I went too far, just look at it. ;)

No changes except one in actual the java code. I renamed a parameter.
I also removed javadoc from some simple getters and setters. I think no comment is needed at a function called `getID()` or `setDate(date foo)`.


src/main/java/de/qabel/core/config/LocalSettings.java
```diff
@@ -14,15 +14,7 @@
...
 public class LocalSettings {
-	/**
-	 * <pre>
-	 *           1     0..*
-	 * LocalSettings ------------------------> LocaleModuleSettings
-	 *           localSettings        &gt;       localeModuleSettings
-	 * </pre>
-	 * Set of module specific local settings
-	 * Field name in serialized json: "module_data"
-	 */
```
Gottox: Which kind of error/warning were produced from this? - Why does that make sense to remove it?

jan-schreib: They don't even appear in the finished javadoc and are there since you created the classes with UML Lab. They have been removed at other places over time now, I just delete everything I saw.

thechauffeur:
Semantically I ACK this PR. We should discuss this style and agree on something regarding stuff like:
* Commenting trivial getter / setter?
* Show relations between classes etc. as a comment?
* ...

After that this PR can be patched / extended or merged.

L-Henke:
I don't think that we need to show the relations created by UMLLab, although they doesn't look too bad in the Javadoc. But they do look bad in the code and I don't think that we really updated the relationships if they have changed, because we only used UMLLab initially and now we just create the class diagram with it.

I further think that if the name of a getter/setter isn't totally descriptive and unambiguous, it should have a Javadoc comment. This often isn't the case.

tl;dr:
Show relationships :-1:
(Trivial) getter/setter comments :+1:

thechauffeur:
> I don't think that we need to show the relations created by UMLLab, although they doesn't look too bad in the Javadoc

As far as I remember these comments which are generated by UML Lab are not Javadoc -- just comments. Think about multi-line comments: `/* ... */` vs. Javadoc comments: `/** ... */`.

I'm also against showing these relationships in the code.

aBothe:
I'm against trivial documentation. I want rational documentation which justifies design/functionality choices, no avoidable redundancy.
I'm also against non-human-readable ASCII-Art in javadoc - does javadoc even allow/recommend  html tags or so?

thechauffeur:
> I further think that if the name of a getter/setter isn't totally descriptive and unambiguous, it should have a Javadoc comment. This often isn't the case.

I think that the whole method signature is totally enough descriptive for a simple getter/setter which does not do magically things like validating and throwing exceptions. I would only document non-trivial things. Are you OK with this @L-Henke?

L-Henke:
@thechauffeur That's pretty much what I said. But in a getter like this https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/storage/PaddedInputStream.java#L63 it is great to have a description, because the unit of the returned ```long``` would have to be guessed without the additional description. 
And there are other getter where this information is missing like this https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/LocalSettings.java#L77. You get no information in which unit the polling interval is returned by looking into the Javadoc. Might be s, ms, ns. Who knows? This is a perfect example of an ambiguous getter which requires a Javadoc comment.

src/main/java/de/qabel/core/config/LocalSettings.java
```diff
@@ -63,24 +55,23 @@ public LocalSettings(long pollInterval, Date dropLastUpdate) {
...
 	 */
 	public void setPollInterval(long value) {
 		this.pollInterval = value;
 	}
 
-	/**
-	 * Returns the poll interval
-	 * @return
+	/*
+	 * @return Returns the poll interval.
```
thechauffeur: In which unit is it returned?

L-Henke: No Idea, but this isn't part of this PR...

thechauffeur:
@L-Henke well that's pretty much what I said :P I would use Javadoc to document your two examples, too. I would **not** document for example: https://github.com/Qabel/qabel-core/pull/279/files#diff-02cbcaeb0ac93a27308fc9c5be258951L77

Also I like things like this:
```Java
/**
 * @return Returns runtime in ms since application start.
 */
public long getRuntime();
```
Instead of
```Java
/**
 * Get the runtime of the application.
 * @return Returns runtime in ms since application start.
 */
public long getTime();
```
So I mean just document non-obviously things and leave out the rest.

L-Henke:
I guess if there are no other opinions, we could add this (maybe with an example) to the code style section of our contributing file.

cburkert:
Since you asked for opinions: I haven't made up my mind on how to handle cases where one of multiple method parameters needs an explaining comment but the others not. Should one add a `@param` only for the one parameter? Wouldn't this look like something has been forgotten? Don't know.

jan-schreib:
Javadoc should be written for all parameters. Not just for one, even if the other is self self-explanatory. Its just incomplete the other way.

jan-schreib:
> @thechauffeur That's pretty much what I said. But in a getter like this https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/storage/PaddedInputStream.java#L63 it is great to have a description, because the unit of the returned long would have to be guessed without the additional description. 
And there are other getter where this information is missing like this https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/LocalSettings.java#L77. You get no information in which unit the polling interval is returned by looking into the Javadoc. Might be s, ms, ns. Who knows? This is a perfect example of an ambiguous getter which requires a Javadoc comment.

@L-Henke in the javadoc you see perfectly what will be returned. You can't see it in the plain comment, though.

thechauffeur:
> Javadoc should be written for all parameters. Not just for one, even if the other is self self-explanatory. Its just incomplete the other way.

Aye.

thechauffeur:
Will commits be added to this? Or in general should this PR be merged now or used later in the whole javadoc vs. documentation clean up?

jan-schreib:
Can be merged now. Should something follow after this, I'll open up a new pullreq.

PR #281, opened 2015-02-09
Modular drop id generation

cburkert:
I implemented a modular drop id generation.
* DropIdGenerator is an abstract drop id generator. Defines API.
* AdjustableDropIdGenerator is a simple randomizing drop id generator as proposed in [drop id choosing discussion](https://github.com/Qabel/qabel-doc/issues/16#issuecomment-52052136)

I think the AdjustableDropIdGenerator is a good placeholder until we come up with something better.
Well it's definitely better than the current fixed dummy id.

### DIFF
--- a/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
+++ b/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
@@ -15,7 +15,7 @@ import java.util.Random;
  * and therefore less anonymity but more performance.
  */
 public class AdjustableDropIdGenerator extends DropIdGenerator {
-	private int usedBits;
+	private final int usedBits;
 
 	/**
 	 * Creates a new random drop id using the whole id namespace.
--- a/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
+++ b/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
@@ -30,7 +30,8 @@ public class AdjustableDropIdGenerator extends DropIdGenerator {
 	 * for the namespace.
 	 * Hence, the collision probability is 1/2^(usedBits).
 	 *
-	 * @param usedBits number of bits used during randomization.
+	 * @param usedBits number of bits used during randomization. Must be in the
+	 *            interval (0, DROP_ID_LENGTH_BYTE*8].
 	 */
 	public AdjustableDropIdGenerator(int usedBits) {
 		if (usedBits <= 0 || usedBits > DROP_ID_LENGTH_BYTE * 8) {
--- a/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
+++ b/src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
@@ -1,0 +1,63 @@
+package de.qabel.core.drop;
+
+import java.math.BigInteger;
+import java.util.Random;
+
+/**
+ * This drop id generator generates drop ids randomly.
+ * It is called adjustable because one can adjust how many bits
+ * of the theoretically available id bit length is used during
+ * the randomization.
+ * This allows a simple tradeoff between collision probability
+ * and communication overhead or in other words between anonymity
+ * and performance.
+ * The rule of thumb is: more utilized bits less collision probability
+ * and therefore less anonymity but more performance.
+ */
+public class AdjustableDropIdGenerator extends DropIdGenerator {
+	private int usedBits;
+
+	/**
+	 * Creates a new random drop id using the whole id namespace.
+	 * Hence, the collision probability is 1/2^(DROP_ID_LENGTH_BYTE*8)
+	 */
+	public AdjustableDropIdGenerator() {
+		this(DROP_ID_LENGTH_BYTE*8);
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


src/test/java/de/qabel/core/drop/DropIdGenerationTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class DropIdGenerationTest {
+	@Rule
+	public ExpectedException exception = ExpectedException.none();
+
+	@Test
+	public void testAdustable() {
```
thechauffeur: adjustable, not adustable. Sorry for nitpicking.

src/main/java/de/qabel/core/drop/DropIdGenerator.java
```diff
@@ -0,0 +1,25 @@
+package de.qabel.core.drop;
+
+import java.util.Arrays;
+
+import org.bouncycastle.util.encoders.UrlBase64;
+
+public abstract class DropIdGenerator {
+	public static final int DROP_ID_LENGTH = 43;
+	public static final int DROP_ID_LENGTH_BYTE = 32;
```
thechauffeur: Why do we want this to be public?

thechauffeur: I know you just used these values as they were already given but since you patch this anyway: These two values depend on each other. It would be cool to set just one value in the code and to calculate the other value.

thechauffeur:
The ´DropURL.checkURL()` method is misleading. But this is not a problem of this PR -- see #283.

I very much like this PR. :+1: for merge.

jan-schreib:
:+1: for merge.

src/main/java/de/qabel/core/drop/DropIdGenerator.java
aBothe: You may need to know which drop id lengths will be accepted when constructing the gen.
- I think it would be an ugly guessing-game if one had to examine IllegalArgumentExceptions thrown in `AdjustableDropIdGenerator(int usedBits)` to negotiate the right size.

src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class AdjustableDropIdGenerator extends DropIdGenerator {
+	private int usedBits;
+
+	/**
+	 * Creates a new random drop id using the whole id namespace.
+	 * Hence, the collision probability is 1/2^(DROP_ID_LENGTH_BYTE*8)
+	 */
+	public AdjustableDropIdGenerator() {
+		this(DROP_ID_LENGTH_BYTE*8);
```
aBothe: I might have missed some documentation about the `8`, but I still would expect to have a more expressive constant name for this value. Perhaps make it `DROP_ID_MAX_LENGTH_BYTE` or similar.

src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+ * of the theoretically available id bit length is used during
+ * the randomization.
+ * This allows a simple tradeoff between collision probability
+ * and communication overhead or in other words between anonymity
+ * and performance.
+ * The rule of thumb is: more utilized bits less collision probability
+ * and therefore less anonymity but more performance.
+ */
+public class AdjustableDropIdGenerator extends DropIdGenerator {
+	private int usedBits;
```
aBothe: If this value is only changed in the constructor, please make it final.

src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
cburkert: The `8` is the number of bits in a byte. BTW the is no maximum or minimum length of drop id in bytes. A drop id has in any case `DROP_ID_LENGTH_BYTE`.

src/main/java/de/qabel/core/drop/DropIdGenerator.java
cburkert: What @aBothe said is exactly the reason why I made them public.

src/main/java/de/qabel/core/drop/AdjustableDropIdGenerator.java
aBothe: Ohhhh, okay, now I got it.
Still, the `@usedBits` param doc should specify that this value must be in the `(0..DROP_ID_LENGTH_BYTE*8]` interval.

cburkert: Ok, I'll add a description to the javadoc

L-Henke:
The adjustable drop id generator is efficient and does its job. This might not be the final solution for creating drop ids, but for the moment this is fine. 

I have no other concerns than the already mentioned by the others. Besides that :+1: for merge.

cburkert:
So, I hopefully fixed all open issues with this PR.

aBothe:
Thanks! :+1:

jan-schreib:
:+1: again.

PR #282, opened 2015-02-09
Changed http status code 404 to 204

L-Henke:
Client side implementation of Qabel/qabel-doc#82
Depends on Qabel/qabel-drop#15

### DIFF
--- a/src/test/java/de/qabel/core/http/DropHTTPTest.java
+++ b/src/test/java/de/qabel/core/http/DropHTTPTest.java
@@ -114,9 +114,9 @@ public class DropHTTPTest {
 		assertEquals(new ArrayList<String>(), response);
 	}
 
-	// GET 404
+	// GET 204
 	@Test
-	public void getRequestForEmptyDropShouldBe404() {
+	public void getRequestForEmptyDropShouldBe204() {
 		// Given
 		DropHTTP dHTTP = new DropHTTP();
 		// When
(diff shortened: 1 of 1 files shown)


thechauffeur:
If Travis passes :+1: for merge.

src/test/java/de/qabel/core/http/DropHTTPTest.java
```diff
@@ -229,7 +229,7 @@ public void shouldBeEmptyWithSinceDate() {
 		int responseCode = dHTTP.head(this.notExistingUrl,
```
aBothe: notExistingUrl? This..should still be 404, right? Also, see the test comment which says `// HEAD 404 + SINCE`

cburkert: No, there is no reason for 404 anymore. If the request is syntactically correct but the drop is emtpy then 204 is the right response, if the request is incorrect then 400 is.
So why did you change the assert but not the comment?

aBothe: 400 stands for bad request, okay. But what if the URL is indeed valid but just not existing? That's 404 iirc.

cburkert: You keep saying "does not exist", but in Qabel/qabel-core#82 we pointed out that a drop has no such property as existence.

aBothe: Okay, not to spoil meta info. Okay, nice then :)

PR #292, opened 2015-02-17
Renamed checkURL() to checkDropIdFromUrl()

### LINKED ISSUE #283: Private checkURL method in DropURL does not check the whole URL
The method instead checks whether the drop id part of the URL is conform to our drop protocol. Hence the name is misleading. Even the Javadoc comment from this method is misleading.

jan-schreib:
Name of the method shouldn't be misleading any more.
Bug was found by @thechauffer in #281.

Closes #283.


jasinai:
:+1:

thechauffeur:
@jan-schreib use GitHubs web interface automagic completion -- you wrote my nick wrong and thus the mention didn't work ;)

PR #293, opened 2015-02-20
Fix erroneous escalation of invalid binary drop messages

cburkert:
It is not unlikely that binary drop messages are received that have the wrong
size compared to expectation implied by the header version byte.
This may for instance happen, if someone maliciously uploads invalid messages or
simply due to broken clients.

This patch fixes the handling of this such unexpected sizes.
They are no longer escalated to a RuntimeException (as was a bug), but are now
logged and then skipped.
Therefore, a new exception class is introduced to distinguish this size
mismatches from other critical version mismatches.


aBothe:
Not sure whether an acceptable state (invalid size) should be handled via exception, but okay, this pr looks good.

thechauffeur:
:+1: for merge.

jan-schreib:
@aBothe I had the same concerns, but its not like a encrypted message that can't be decrypted for whatever reason. That obviously is not a case for an exception.

:+1 for merge and go.

PR #300, opened 2015-03-03
Noise integration

### LINKED ISSUE #290: Merge the noise implementation
We implemented a new cryptographic core to encrypt drop messages. This implementation needs to be incorporated into qabel-core and the documentation (and maybe some examples) need to be updated.
@L-Henke, @roeslpa 

(See Qabel/qabel-doc/issues/72 for some of the reason for the new implementation.)

L-Henke:
Closes #290 
~~This is an early state of the noise integration.~~

Some concepts are currently bad and will be changed. ~~This PR is only to save my work and to ask other developers on details for required refactorizations. It is also not perfectly cleaned up.~~

A concept for a platform independent integration or even compilation of the ```.so``` is completely missing.

~~Only check this PR if you are interested in the new concept.~~

Edit: Current goal was to pass all tests, not to make a totally optimized concept.


L-Henke:
@f-porter could you have a look if the changes in the TypeAdapters are fine? They are working, but I'm not sure if the design is perfekt. 
https://github.com/L-Henke/qabel-core/blob/noiseMerge/src/main/java/de/qabel/core/config/ContactTypeAdapter.java
https://github.com/L-Henke/qabel-core/blob/noiseMerge/src/main/java/de/qabel/core/config/IdentitiesTypeAdapter.java

L-Henke:
I guess this PR is now ready to have a look at and discuss how to improve the currently rough edged. These are at least:

* Platform independent integration of JNI classes
* ```AbstractBinaryDropMessage``` and ```BinaryDropMessageV0```
* Reaction on ```QblSpoofedSenderException``` in ```DropController```

Since this PR touches a lot of different parts of ```qabel-core``` I would suggest, that people can also ack selected parts of the PR if they want to.

thechauffeur:
@L-Henke is totally right that specific persons of you @Qabel/developer should ack specfic parts. Some examples are mentioned above.

f-porter:
The changes in the TypeAdapter look fine for now. Optimizations of the design should be part of a refactoring of the config TypeAdapters, anyway.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,67 @@
...
+												jobject obj,
+												jbyteArray n, jbyteArray p)
+{
+	jboolean qIsCopy, pIsCopy;
+	jbyte *_n, *_result, *_p;
+	jbyteArray result;
+	jsize length;
+
+	length = (*env)->GetArrayLength(env, n);
+	_result = (jbyte *) malloc(length * sizeof(jbyte));
```
jan-schreib: 1. There is no need to cast the return value of malloc.
2. A check if the malloc has succeeded would be nice.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,67 @@
+#include <jni.h>
+#include <stdlib.h>
+
+#include "curve25519-jni.h"
+#include "curve25519.h"
+
+JNIEXPORT jbyteArray JNICALL
+Java_de_qabel_core_crypto_Curve25519_cryptoScalarmult(JNIEnv * env,
+												jobject obj,
+												jbyteArray n, jbyteArray p)
```
jan-schreib: This is a little too much of intendation.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,67 @@
...
+	(*env)->SetByteArrayRegion(env, result, 0, length, (jbyte *) _result);
+	free(_result);
+
+	return (result);
+}
+
+JNIEXPORT jbyteArray JNICALL
+Java_de_qabel_core_crypto_Curve25519_cryptoScalarmultBase(JNIEnv * env,
+													jobject obj,
+													jbyteArray n)
```
jan-schreib: Indentation as mentioned above.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,67 @@
...
+{
+	jboolean qIsCopy;
+	jbyte *_n, *_result;
+	jbyteArray result;
+	jsize length;
+
+	length = (*env)->GetArrayLength(env, n);
+	_result = (jbyte *) malloc(length * sizeof(jbyte));
```
jan-schreib: Same as above.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,67 @@
...
+	result = (*env)->NewByteArray(env, length);
+
+	_n = (jbyte *) (*env)->GetByteArrayElements(env, n, &qIsCopy);
+
+	crypto_scalarmult_base((unsigned char *) _result,
+						   (unsigned char *) _n);
```
jan-schreib: Indentation.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+	if (qIsCopy) {
+		(*env)->ReleaseByteArrayElements(env, n, _n, JNI_ABORT);
+	}
+	if (pIsCopy) {
+		(*env)->ReleaseByteArrayElements(env, p, _p, JNI_ABORT);
+	}
+
+	(*env)->SetByteArrayRegion(env, result, 0, length, _result);
+
+	return (result);
```
jan-schreib: No brackets here, pls.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+
+	crypto_scalarmult_base((unsigned char *) _result, (unsigned char *) _n);
+
+	if (qIsCopy) {
+		(*env)->ReleaseByteArrayElements(env, n, _n, JNI_ABORT);
+	}
+
+	(*env)->SetByteArrayRegion(env, result, 0, length, _result);
+
+	return (result);
```
jan-schreib: See above.

jni/curve25519-jni.h
```diff
@@ -0,0 +1,29 @@
+/* DO NOT EDIT THIS FILE - it is machine generated */
+#include <jni.h>
+/* Header for class de_qabel_core_crypto_Curve25519 */
+
+#ifndef _Included_de_qabel_core_crypto_Curve25519
+#define _Included_de_qabel_core_crypto_Curve25519
+#ifdef __cplusplus
```
jan-schreib: We don't use c++, the ifdefs for c++ can be removed.

L-Henke: As you can read in the diff you have commented: ```* DO NOT EDIT THIS FILE - it is machine generated */``` :)

jan-schreib: Never trust machines. They will kill us in the end.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+	jbyte *_n, *_p;
+	jbyteArray result;
+	jsize length;
+
+	length = (*env)->GetArrayLength(env, n);
+	jbyte _result[length];
+	result = (*env)->NewByteArray(env, length);
+
+	_n = (jbyte *) (*env)->GetByteArrayElements(env, n, &qIsCopy);
+	_p = (jbyte *) (*env)->GetByteArrayElements(env, p, &pIsCopy);
```
jan-schreib: No casting needed here. If some person would've read the JNI documentation instead of copying code... ah nevermind.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+	jboolean qIsCopy;
+	jbyte *_n;
+	jbyteArray result;
+	jsize length;
+
+	length = (*env)->GetArrayLength(env, n);
+	jbyte _result[length];
+	result = (*env)->NewByteArray(env, length);
+
+	_n = (jbyte *) (*env)->GetByteArrayElements(env, n, &qIsCopy);
```
jan-schreib: No casting needed her either.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+	_p = (jbyte *) (*env)->GetByteArrayElements(env, p, &pIsCopy);
+
+	crypto_scalarmult((unsigned char *) _result, (unsigned char *) _n, (unsigned char *) _p);
+
+	if (qIsCopy) {
+		(*env)->ReleaseByteArrayElements(env, n, _n, JNI_ABORT);
```
jan-schreib: I'm not sure about using JNI_ABORT here. The [documentation](http://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/functions.html#wp17314) says that 0 is what most developers want. Maybe @Gottox, @L-Henke could look at this, too.

jni/curve25519-jni.c
```diff
@@ -0,0 +1,59 @@
...
+
+	length = (*env)->GetArrayLength(env, n);
+	jbyte _result[length];
+	result = (*env)->NewByteArray(env, length);
+
+	_n = (jbyte *) (*env)->GetByteArrayElements(env, n, &qIsCopy);
+
+	crypto_scalarmult_base((unsigned char *) _result, (unsigned char *) _n);
+
+	if (qIsCopy) {
```
jan-schreib: Things I mentioned above apply to this function, too.

src/main/java/de/qabel/core/drop/DropController.java
```diff
@@ -232,17 +243,21 @@ public DropResult send(DropMessage<? extends ModelObject> message, Collection<Co
...
 				continue;
 			}
-			for (Contact c : contacts) {
-				DropMessage<?> dropMessage = binMessage.disassembleMessageFrom(c);
+			for (Identity identity : identities) {
+				DropMessage<?> dropMessage = null;
+				try {
+					dropMessage = binMessage.disassembleMessage(identity);
+				} catch (QblSpoofedSenderException e) {
+					break;
```
jan-schreib: Since we can't notify a potential user about the spoofing attempt right now, a TODO would be nice.
@L-Henke and I already talked about this.

jni/curve25519-jni.c
Gottox: If I understand this piece of code correctly, this ``ReleaseByteArrayElements`` call will free "C-Space-Memory" without coping the elements back to the Java world. Which seems okay to. Me. @L-Henke can you explain this a little more?

L-Henke: I currently cant, because the C code is from a Qabel developer who isn't at the company anymore. But I'll have a further look at that C code.

src/main/java/de/qabel/core/config/Contact.java
```diff
@@ -31,17 +31,17 @@
 	 * @return QblPrimaryPublicKey
```
jasinai: It returns QblECPublicKey

src/main/java/de/qabel/core/config/Identity.java
```diff
@@ -61,14 +61,14 @@ public void setPrimaryKeyPair(QblPrimaryKeyPair key)
 	 * Returns the primary key pair of the identity
 	 * @return QblPrimaryKeyPair
```
jasinai: It returns QblECKeyPair

jni/curve25519.c
```diff
@@ -0,0 +1,160 @@
+#include "curve25519.h"
```
aBothe: No copyright/copyleft/license header required?
I probably just didn't read any readme, but where did you take this file from? Any url so we may diff the original document against this?
This code is a little too critical to be pasted in without any meta info where it came from, imho.

src/main/java/de/qabel/core/crypto/QblECPublicKey.java
```diff
@@ -0,0 +1,50 @@
+package de.qabel.core.crypto;
+
+import org.bouncycastle.util.encoders.Hex;
+
+import java.util.Arrays;
+
+public class QblECPublicKey {
```
aBothe: Please add at least 'Elliptic Curve Public Key' to its javadoc.

jni/curve25519.c
schulze: The curve25519 code was extracted from the public domain [TweetNaCl](http://tweetnacl.cr.yp.to) code: http://tweetnacl.cr.yp.to/20140427/tweetnacl.c. Only a few of the function names where changed.
Technically we don't need to add a license header, but we should still acknowledge the source.

aBothe:
Could you update the code pls?
Are there any further obstacles why this can't be merged yet?
Remember, it's continous integration, nothing has to be absolutely completed/perfect to be merged, imho.

L-Henke:
According to our guidelines, the code only gets updated when its ready to merge. We have decided to do this to avoid unnecessary work.

aBothe:
Sure okay.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -920,4 +289,268 @@ public boolean decryptFileAuthenticatedSymmetricAndValidateTag(InputStream input
...
+	 * @return derived key or null on unexpected errors
+	 * @throws java.security.InvalidKeyException if secret cannot be used as a HMAC secret
+	 */
+	byte[] kdf(byte[] secret, byte[] extraSecret, byte[] info, int outputLen) throws InvalidKeyException {
+		ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+		byte[] t = new byte[H_LEN];
+
+		// Not required to init the ByteArrayOutputStream with the
+		// expected length, but it might improve the performance
+		ByteArrayOutputStream bs = new ByteArrayOutputStream(info.length + 1 + 32 + extraSecret.length);
```
cburkert: What does the literals 32 and 1 stand for?

L-Henke: 1 byte  for ```c``` and 32 bytes for  ```t```. Of course it wouldn't be necessary to init the output stream with the size of the input.

cburkert: There is a constant for the size of t: H_LEN.
BTW: Great variable naming ;-)

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -34,37 +31,40 @@
...
+	private final static int SYMM_NONCE_SIZE_BYTE = 12;
 	private final static int AES_KEY_SIZE_BYTE = 32;
 	private final static int AES_KEY_SIZE_BIT = AES_KEY_SIZE_BYTE * 8;
-	final static int ENCRYPTED_AES_KEY_SIZE_BYTE = 256;
+
+	private static byte[] SUITE_NAME = "Noise255/AES256-GCM\0\0\0\0\0".getBytes();
+	private static final String HMAC_ALGORITHM = "HMac/SHA512";
+	private static final int H_LEN = 64;
+	private static final int CV_LEN_BYTE = 48;
+	private static final int SYMM_KEY_LEN_BYTE = 32;
```
cburkert: What's the difference to AES_KEY_SIZE_BYTE?

src/main/java/de/qabel/core/crypto/CryptoUtils.java
L-Henke: Please direct your complains to the inventor of the protocol. https://github.com/trevp/noise/wiki/Boxes#key-derivation I'm not going to invent names for these variables. I further didn't use ```H_LEN``` for initialization because is wasn't at the write function in the pseudo code, although it is available (```t[0:32]```).

L-Henke: It further isn't H_LEN, it is just 32. No matter what H_LEN is.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
L-Henke: Currently there is non other than keeping the noise key size and the file encryption key size independent.

L-Henke: A little more detailed: You could change ```AES_KEY_SIZE_BYTE``` to 16 and file encryption would work just fine. But changing ```SYMM_KEY_LEN_BYTE``` to 16 would break our noise implementation, because it's a different protocol with other values (e.g. ```SUITE_NAME = "Noise255/AES256-GCM\0\0\0\0\0"``` would have to be different). We haven't implemented any other box type than ```Noise255/AES-256-GCM```.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
```diff
@@ -920,4 +289,268 @@ public boolean decryptFileAuthenticatedSymmetricAndValidateTag(InputStream input
...
+	 * @throws java.security.InvalidKeyException if secret cannot be used as a HMAC secret
+	 */
+	byte[] kdf(byte[] secret, byte[] extraSecret, byte[] info, int outputLen) throws InvalidKeyException {
+		ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+		byte[] t = new byte[H_LEN];
+
+		// Not required to init the ByteArrayOutputStream with the
+		// expected length, but it might improve the performance
+		ByteArrayOutputStream bs = new ByteArrayOutputStream(info.length + 1 + 32 + extraSecret.length);
+		for (int c = 0; c <= Math.ceil((double) outputLen / H_LEN) - 1; c++){
```
cburkert: The for loop condition is really unintuitive with this <= and minus 1 (which compensates each other).
I'd suggest to change it to:

     c < Math.ceil((double) outputLen / H_LEN)

or old school without Math:

      c < (outputLen + H_LEN - 1) / H_LEN

src/main/java/de/qabel/core/crypto/CryptoUtils.java
cburkert: Ok, thanks.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
L-Henke: This again has been done to be close to the pseudo code ```for c = 0...(ceil(output_len / H_LEN) - 1)```, but I think this probably can be changed.

src/main/java/de/qabel/core/crypto/CryptoUtils.java
cburkert: Oh, I see, H_LEN is 64. Maybe the 32 is H_LEN / 2?
But I understand: It's a magic constant from the noise pseudo code :-)

aBothe: Could you still mention from which codefile/method/line/doc these numbers come from in-code? 
Mind the Bus-Factor :smile:

src/main/java/de/qabel/core/drop/DropController.java
```diff
@@ -198,15 +207,17 @@ public DropResult send(DropMessage<? extends ModelObject> message, Collection<Co
 	 * Retrieves a drop message from given URL
 	 *
 	 * @param url      URL where to retrieve the drop from
-	 * @param contacts Contacts to check the signature with
+	 * @param identities Identities to decrypt message with
+	 * @param contacts Contacts to search sender in
 	 * @return Retrieved, encrypted Dropmessages.
 	 */
-	public Collection<DropMessage<?>> retrieve(URL url, Collection<Contact> contacts) {
+	public Collection<DropMessage<?>> retrieve(URL url, Collection<Identity> identities,
```
cburkert: Nothing important, but this line wrap is just hilarious.

src/main/java/de/qabel/core/drop/DropController.java
```diff
@@ -232,17 +243,22 @@ public DropResult send(DropMessage<? extends ModelObject> message, Collection<Co
...
-						break;
+					for (Contact c : contacts) {
+						if (c.getKeyIdentifier().equals(dropMessage.getSenderKeyId())){
+							if (dropMessage.registerSender(c)){
+								plainMessages.add(dropMessage);
+							}
+						}
 					}
-					plainMessages.add(dropMessage);
 					break; // sender found for this message
```
cburkert: The comment doesn't make sense anymore, because we don't search the sender with noise.

cburkert:
I reviewed all but the JNI code and couldn't find any real issues.
:+1: for merge from me

src/main/java/de/qabel/core/drop/DropController.java
L-Henke: It doesn't look that way in my IDE. There the second line is aligned with the ```(```.

src/main/java/de/qabel/core/drop/DropController.java
L-Henke: Does the ```break``` even make sense at that point? Wouldn't it have to be moved with ```plainMessages.add(dropMessage);```

cburkert: Then it would only break out of the inner for loop, iterating over the contacts. One could put an additional break there.

src/main/java/de/qabel/core/drop/DropController.java
cburkert: Indeed, strange github... Never mind then.

src/main/java/de/qabel/core/crypto/BinaryDropMessageV0.java
```diff
@@ -65,72 +56,40 @@ int getPayloadSize() {
...
-			rawPlainText = cu.decryptSymmetric(encMessage, new SecretKeySpec(
-					rawAesKey, CryptoUtils.SYMM_KEY_ALGORITHM));
+			decryptedPlaintext = cu.readBox(identity.getPrimaryKeyPair(),
+					Arrays.copyOfRange(binaryMessage, HEADER_SIZE, binaryMessage.length));
 		} catch (InvalidKeyException e) {
-			logger.debug("Invalid AES key");
-			return null;
+			logger.debug("Message not meant for this recipient");
+		} catch (InvalidCipherTextException e) {
+			logger.debug("Message not meant for this recipient");
```
cburkert: I guess this log message is wrong, because looking at readBox, InvalidCipherTextException can have many reasons not related to the recipient. InvalidCipherTextException also provides a handy error description string which should be logged, too.

jan-schreib:
I changed the JNI code. I didn't see any more issues. For now I think this is ready to be merged.

L-Henke:
Is it possible to throw an exception from there? That case should really not occur in a normal program execution. (http://www.rgagnon.com/javadetails/java-0323.html)

jan-schreib:
I will add a throw there. Haven't thought of that... thx @L-Henke

aBothe:
Congrats, thanks and kudos for having this merged! :D

PR #304, opened 2015-03-10
Fixes #302 Added guideline: Should use Java 7 features

### LINKED ISSUE #302: Java 7 usage of shortened type brackets
Although we could use the shortened type brackets from Java 7 e.g. ```ArrayList<Integer> al = new ArrayList<>``` instead of ```ArrayList<Integer> = new ArrayList<Integer>```, we often don't use it. I think we should do this the same way in the code. 
https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/drop/DropController.java#L31

thechauffeur:


CONTRIBUTING.md
```diff
@@ -65,6 +65,7 @@ Regarding the look and feel most -- but not all -- of this is reflected by the E
 * Use tabs instead of spaces.
 * Try to fit your code into **80** characters but do not wrap your line of code just for the sake of wrapping it. Use **120** in your auto format settings.
 * Statements like ```if``` always with ```{ }```. Do not just write ```if (foo) bar;```
+* Our code does not need to be compitable to versions below Java 7. That means you can and should use Java 7 features and syntax.
```
aBothe: compatible or compilable?

jan-schreib: @aBothe is right, it should be `compatible`.

jan-schreib:
When the typo is corrected I ack this.

Issue #298, opened 2015-03-02
Port Config and Contact components to ackack


thechauffeur:
@f-porter wants to look into how to use ackack using #252 (which references all example pull requests) and ask @Gottox open questions if needed. This is because we need to solve the problem how to get / change resources like settings and contacts -- which is described in #185.

A pull request to propose how to solve the problem described above is the next step.

Issue #303, opened 2015-03-09
Define how to deal with dependent method calls


L-Henke:
We sometimes have dependent method calls. (e.g. one method have to be called after another).

E.g: ```registerSender``` has to be called to complete the serialization of a Drop message:
https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/drop/DropMessage.java#L74

And:
The serialize and deserialize of the ```AbstractBinaryDropMessage``` are related to which constructor has been used ```byte[] binaryMessage``` or ```DropMessage<?> dropMessage```. These methods are not testing if the current call is actually possible.

How should be deal with that kind of dependent methods. Is a Javadoc comment enough, or is there maybe a better way to do this? Should these methods check if their call is valid in the current context?

L-Henke:
#316 has a dependent method call. The Persistence class can be only used after ```init("password")``` has been called. All methods are checking if ```init``` has been called and will throw an ```IllegalStateException``` with an explanation why this call has failed if ```init``` hasn't been called before. 

I think this is how we should deal with dependent method calls. This is a lot better than checking why an unspecified error has occurred. 

@Qabel/developer please let me know if you can approve this concept as our general approach.

cburkert:
Sounds valid to me.

L-Henke:
We want to have this concept and will document it.

Issue #287, opened 2015-02-10
Use Javadoc to document our software architecture


thechauffeur:
not our documentation repository.

See Qabel/qabel-doc#83

@jan-schreib, @jasinai and myself will work on this.

thechauffeur:
@jan-schreib started to re-work our Javadoc documentation in #279.

thechauffeur:
We need to update the CONTRIBUTING file because of things mentioned in #279.

Later we should revise the documentation / spec, again.

thechauffeur:
* [ ] Add Javadoc to https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/EntityMap.java especially after #305 is merged.

thechauffeur:
@cburkert please add Javadoc to the class mentioned above and let the PR for that close this issue.

PR #305, opened 2015-03-11
Issue298 part1 port contact component to ackack

f-porter:
Part one of #298.
I am not sure if the design/API of the ContactsActor is perfect - so please have a look and give feedback.

Please note: Javadoc is still missing as the API of the ContactsActor could be changed prior to merge, I will add it in a later commit.


f-porter:
@Gottox, do you have any idea why the build failed with:
```Project with path ':ackack' could not be found in root project 'qabel-core'.```

**Comment by Gottox** [comment]
Sure, that's rather easy to fix, but there's another issue which will appear afterwards.

**Comment by Gottox** [comment]
The problem is, that the buildserver doesn't use the qabel repository to test but the qabel-core repository. It's missing the ackack dependency.

workaround is to add ackack as a git submodule to qabel-core.

**Comment by f-porter** [comment]
okay, I expected something like that.
So shall I add it as submodule or how did you do that in #272?

**Comment by Gottox** [comment]
Yes, as long as ackack is not distributed through maven.

**Comment by thechauffeur** [comment]
@Qabel/developer you should all review this. @Gottox is assignee for special stuff.

**Comment by L-Henke** [comment]
Couldn't find any major flaws. The concept seems to be correct, but I can't make a fully qualified statement about that. The tests seems to cover any kind of operations on contacts.

File: src/main/java/de/qabel/core/config/Contact.java [hunk 1]
```diff
@@ -91,6 +95,14 @@ protected Contact(String ownerKeyId, Collection<DropURL> dropUrls, QblPrimaryPub
 		this.setPrimaryPublicKey(pubKey);
 	}
 
+	/**
+	 * Creates and instance of Contact without any attributes set
```
jasinai: Creates **an** instance?

f-porter: sure

src/main/java/de/qabel/core/config/EntityMap.java
```diff
@@ -12,7 +13,11 @@
  * 
  * @see Entity
  */
-abstract public class EntityMap<T extends Entity> {
+abstract public class EntityMap<T extends Entity> implements Serializable {
+	/**
+	 *
+	 */
```
cburkert: Can you please remove these empty javadocs above the serial numbers.

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -0,0 +1,103 @@
...
+		}
+		return defaultContactsActor;
+	}
+
+	private static final String WRITE_CONTACTS = "writeContacts";
+	private static final String RETRIEVE_CONTACTS = "retrieveContacts";
+	private static final String REMOVE_CONTACTS = "removeContacts";
+
+	public ContactsActor (Contacts contacts) {
+		ContactsActor.contacts = contacts;
```
cburkert: This constructor overrides the package private static variable contacts for *all* instances of ContactsActor.
I don't understand why you need this.
Generally, static variables should not be initialised in an instance context like a constructor.

src/main/java/de/qabel/core/config/EntityMap.java
f-porter: Done

src/main/java/de/qabel/core/config/ContactsActor.java
f-porter: Yes, that was a mistake. The constructor should be for testing purposes to be able to use another instance of Contacts than the default one.

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -0,0 +1,104 @@
...
+import de.qabel.ackack.Actor;
+import de.qabel.ackack.MessageInfo;
+import de.qabel.ackack.Responsible;
+
+/**
+ * Actor which handles the access to contacts of Qabel.
+ *
+ */
+public class ContactsActor extends Actor {
+	static Contacts defaultContacts;
```
L-Henke: The ```defaultContacts``` aren't used for anything useful anymore. It's just passed in the static constructor and has no further use.

f-porter: Yes, I know. We haven't decided how to set the contacts for the defaultContactsActor (initializing qabel, deserializing contacts,...), so defaultContacts are not set by any method yet. I would add the initialization of defaultContacts later, when this is clear. Does this make sense to you or would you do this in another way?

jan-schreib:
@f-porter please updated this pullreq to the current master and we will continue to review it. Thanks.

f-porter:
Done. Have a look.

ackack
```diff
@@ -1 +1 @@
-Subproject commit 8a43daca64782873f13bb853f11c10bedd1e9652
+Subproject commit 5fa8b255dedf6f1e1c955f48ac66cccbe963bce8
```
Gottox: Can you update to current master?

src/main/java/de/qabel/core/config/ContactsActor.java
L-Henke: The ```defaultContacts``` are not initialized, so they are just ```null```. You could just directly initialize the contacts actor with ```null``` and change it when we know how to get the default contacts. But this field currently doesn't have any benefit. 

The initialization of the default contacts has to be somehow in the ```getDefault``` method anyway, because otherwise the default ContactActor wouldn't work without a prior initialization of the ```defaultContacts``` field, which would make the method pretty useless. 

So I would just remove the field for now and take care of the initialization later.

ackack
f-porter: done

f-porter:
The travis build failed with:
```Exception in thread "main" javax.net.ssl.SSLHandshakeException: java.security.cert.CertificateException: No subject alternative DNS name matching services.gradle.org found.```
Seems like I have to re-trigger the build, but I don't find the button to do so...?

**Comment by jan-schreib** [comment]
@f-porter I re-triggered the test a couple hours ago and everything passed.

File: src/main/java/de/qabel/core/config/ContactsActor.java (hunk 4 above)
Comment (f-porter) [inline src/main/java/de/qabel/core/config/ContactsActor.java]: Okay, but then I should delete the ```getDefault``` method as well, as it does not have any functionality right now, shouldn't I?

Comment (L-Henke) [inline src/main/java/de/qabel/core/config/ContactsActor.java]: Why wouldn't it have any functionality? You can get the default contacts actor from it, it will just have no contacts added when the first ```getDefault``` is called. But of course one can add contacts and another caller would get the contacts actor with these contacts.

And when we have a way to persist and load data, the default contacts actor can contain the saved contacts. I don't see any need to delete that method.

Comment (f-porter) [inline src/main/java/de/qabel/core/config/ContactsActor.java]: Yes, that sounds reasonable.

File: src/main/java/de/qabel/core/config/ContactsActor.java [hunk 6]
```diff
@@ -0,0 +1,103 @@
...
+		return post(info, (Serializable[]) keyIdentifiers);
+	}
+
+	@Override
+	protected void react(MessageInfo info, Object... data) {
+		switch(info.getType()) {
+		case WRITE_CONTACTS:
+			Contact contact;
+			for (int i = data.length-1; i>=0; i--) {
+				contact = (Contact) data[i];
```
L-Henke: Since ```data``` might be every object, you might check with ```instanceof```if the object can be casted to a Contact.

f-porter: Actually the type is checked by the call of method ```writeContacts(final Contact... contacts)```.

thechauffeur: IIRC we discussed that this should be done. Correct me if I'm wrong.

src/main/java/de/qabel/core/config/Contact.java
```diff
@@ -91,6 +92,14 @@ protected Contact(String ownerKeyId, Collection<DropURL> dropUrls, QblECPublicKe
 		this.setEcPublicKey(pubKey);
 	}
 
+	/**
+	 * Creates an instance of Contact without any attributes set
+	 * Attention: This constructor is intended for deserialization purposes when getting copied by ContactsActor
```
thechauffeur: According to #303 in cases where dependencies like this are existing, every method need to check for the correct state of the object and throw an `IllegalStateException` otherwise. This is also true -- maybe even more -- for the other protected constructor but this one is not part of this pull request.

Correct me if I'm wrong.

Either do this or open an issue so that this problem is addressed after merging this PR.

src/main/java/de/qabel/core/config/EntityMap.java
```diff
@@ -35,10 +36,19 @@ public boolean add(T entity) {
 		}
 	}
 
+	public boolean replace(T entity) {
```
thechauffeur: Javadoc would be nice. E.g. this method works just like the method `boolean add(T entity)` if `entity` is not yet contained. That said the other methods here could need Javadoc, too. I will this task to #287.

src/main/java/de/qabel/core/config/Contact.java
thechauffeur: Well I don't see any dependencies which would cause things like null pointer exceptions in methods of this class. Nevertheless this could be done better and I added issue #323 for this -- it can be done later.

src/main/java/de/qabel/core/config/ContactsActor.java
thechauffeur: This can be done later so I added an issue for that #324

src/main/java/de/qabel/core/config/EntityMap.java
thechauffeur: This means nothing need to be done in this PR.

thechauffeur:
Ack.

thechauffeur:
:+1:

PR #306, opened 2015-03-30
silence endless outputs

jan-schreib:


aBothe:
Logs look less messy now despite they somehow still fail. Why do they fail? Will an already open PR fix this?

jasinai:
:+1:

PR #316, opened 2015-04-22
Added object persistence

### LINKED ISSUE #18: Platform indepentent object persistance
We need a way to persist object on Android and PC without Android specific code. Maybe @Gottox know more about the persistence possibilities on Android and how this can be done platform independent.

L-Henke:
This PR is based on #305.

This adds a Persistence class to persist Serializable objects. It probably isn't completely finished, but worth a look. SQLitePersistence uses SQLite as a simple blob storage. 

Entity has a new UUID field which can be used to get or remove a specific persisted object. I introduced this artificial ID, because it is stored unencrypted in the database (to allow lookups). Using actual values like the key id would leak this information.

The serialized data is encrypted with AES256 GCM and a random nonce per object. A master key is used to encrypt all objects. This master key itself is also AES256 GCM encrypted with nonce and a key derived from via PBKDF2 from a password and a random salt. This allows easy password changed by just re-encrypting the master key instead of re-encrypting all stored values.

The UUID is provided to the GCM encryption as authenticated data. This detects if somebody switches the encrypted object with another encrypted object.

The ContactActor has been modified to store, delete or get Contacts from the Persistence class.

A simple test has been added, but more tests are required and will follow (password change, deletion, etc.)

How the persistence layer should receive the password has to be determined. Currently, the ContactActor will provide it to the persistence class. This will not be the final concept.

Closes #18 

TODO:
- [x] Add further tests

### DIFF
--- a/src/main/java/de/qabel/core/config/Persistence.java
+++ b/src/main/java/de/qabel/core/config/Persistence.java
@@ -48,7 +48,7 @@ public abstract class Persistence {
 	/**
 	 * Initializes the persistence class
 	 */
-	final void init(String password) {
+	final void init(char[] password) {
 		this.keyParameter = getMasterKey(deriveKey(password, getSalt(false)));
 		initCalled = true;
 	}
@@ -68,7 +68,7 @@ public abstract class Persistence {
 	 * @param newPassword New password
 	 * @return True if password has been changed
 	 */
-	final boolean changePassword(String oldPassword, String newPassword) {
+	final boolean changePassword(char[] oldPassword, char[] newPassword) {
 		return reEncryptMasterKey(deriveKey(oldPassword, getSalt(false)), deriveKey(newPassword, getSalt(true)));
 	}
 
@@ -144,12 +144,12 @@ public abstract class Persistence {
 	 * @param salt Salt for key derivation
 	 * @return Encryption key for serialization/deserialization
 	 */
-	private KeyParameter deriveKey(String password, byte[] salt) {
+	private KeyParameter deriveKey(char[] password, byte[] salt) {
 		SecretKey key;
 		try {
 			SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance(SECRET_KEY_ALGORITHM);
 			key = secretKeyFactory.generateSecret(
-					new PBEKeySpec(password.toCharArray(), salt,
+					new PBEKeySpec(password, salt,
 							PBKDF2_ROUNDS, AES_KEY_SIZE_BIT));
 			return new KeyParameter(key.getEncoded());
 		} catch (InvalidKeySpecException e) {
--- a/src/main/java/de/qabel/core/config/Persistence.java
+++ b/src/main/java/de/qabel/core/config/Persistence.java
@@ -43,7 +43,10 @@ public abstract class Persistence<T> {
 			logger.fatal("Cannot find selected algorithm!", e);
 			throw new RuntimeException("Cannot find selected algorithm!", e);
 		}
-		connect(dbName);
+		if (!connect(dbName)) {
+			logger.fatal("Cannot connect to database!");
+			throw new RuntimeException("Cannot connect to database!");
+		}
 		this.keyParameter = getMasterKey(deriveKey(PASSWORD, getSalt(false)));
 	}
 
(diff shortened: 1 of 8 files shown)


jan-schreib:
I've red a couple of lines and noticed that you are using ```Strings``` for passwords.
```Strings``` are immutable in Java. Nobody knows how long they will be in the string pool (and memory) and nobody can zero out the content. Trying to zero the password string would probably make it worse.
I highly recommend to use char arrays for passwords and to clean the content afterwards.
See [this](http://www.oracle.com/technetwork/java/seccodeguide-139067.html#2) for more.

**Comment by L-Henke** [comment]
Currently this shouldn't be an issue, because the derived key will stay in memory the whole runtime and the password methods will be completely removed anyway. It's just a workaround until we know how to provide the password to the persistence class. But I changed it anyway.

```private final static char[] encryptionPassword = "qabel".toCharArray();``` I guess this string is always stays in memory, but it's just for the tests.

File: src/main/java/de/qabel/core/config/Persistence.java [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+
+	public Persistence(String dbName) {
+		this.cryptoutils = new CryptoUtils();
+		try {
+			this.secretKeyFactory = SecretKeyFactory.getInstance(SECRET_KEY_ALGORITHM);
+		} catch (NoSuchAlgorithmException e) {
+			logger.fatal("Cannot find selected algorithm!", e);
+			throw new RuntimeException("Cannot find selected algorithm!", e);
+		}
+		connect(dbName);
```
jan-schreib: The return type is not checked. In this case I think it is important. The exceptions are logged, but do not have any consequences when the return type is ignored.

L-Henke: I know :) Also handling over the db name as a String is not very type safe. Will get optimized..

thechauffeur:
@L-Henke since we talked about how to handle dependent method calls I wondered what your newest (commit)[https://github.com/L-Henke/qabel-core/commit/752a32b5f69f41be2768bfafcdfafb53ddb8f06d] is about. Do you not need to check for the initialization anymore? In the commit message it says that this commit is a workaround until $something. Why do you introduce it? Please explain this a bit.

L-Henke:
@cburkert and I figured out, that the Persistence class could be implemented without the init method and the resulting state. But we are (or at least I'm) unsure which concept is better. Now a member of the sub-class ```SQLitePersistence``` is not (directly) initialized by its constructor. To fully understand this, I'll try to show the construction process of the ```SQLitePersistence```.

* ```new SQLitePersistence(dbName)```.
* ```super(dbName)```
* The ```Persistence``` constructor now calls the abstract ```connect(dbName)``` method. 
* The concrete ```connect(dbName)``` method of ```SQLitePersistence``` is called.
* ```connect(dbName)``` initialized the connection and sets the member ```connection```

As a result, ```connection``` is initialized by a method invoked by the super constructor. Further the db name has to be handed over to the super constructor which has to hand it back into the ```connect``` method. Also the type of ```dbName``` has to be some supertype which fits all possible database types, because it is handed over from the super class to the sub class. 

I was going to discuss with the other developers which concept they are favoring.

src/main/java/de/qabel/core/config/Persistence.java
```diff
@@ -0,0 +1,216 @@
...
+		}
+		return reEncryptMasterKey(deriveKey(oldPassword, getSalt(false)), deriveKey(newPassword, getSalt(true)));
+	}
+
+	/**
+	 * Connect to the used database
+	 * @param dbName Database name to connect to
+	 * @return Result of the operation
+	 */
+	abstract boolean connect(String dbName);
```
thechauffeur: Use a generic here -- you can give the type to the Persistence constructor in the sub class.

thechauffeur:
I reviewed the new commits -- :+1:

src/test/java/de/qabel/core/config/PersistenceTest.java
```diff
@@ -0,0 +1,200 @@
...
+	public void getEntitiesTest() {
+		PersistenceTestObject pto = new PersistenceTestObject("pto");
+		PersistenceTestObject pto2 = new PersistenceTestObject("pto2");
+
+		Assert.assertTrue(persistence.persistEntity("1", pto));
+		Assert.assertTrue(persistence.persistEntity("2", pto2));
+
+		List<Object> objects = persistence.getEntities(PersistenceTestObject.class);
+		System.out.println(objects.size());
+		Assert.assertTrue(objects.size() == 2);
```
jasinai: Can you remove the System.out?
And can you change the assertTrue to assertEquals?

jasinai:
Ack :+1:

Issue #313, opened 2015-04-13
Automatically send fake drop messages


thechauffeur:

thechauffeur:
During #317 we noticed what will need for this:
* [ ] Using noise random padding for this
* [ ] A way to get all drop servers
* [ ] Generating random drop URLs
* [ ] We need mocks / stubs to find out where a fake message (random drop URL) has been retrieved to actually test this functionality.

cburkert:
The `de.qabel.core.drop.DropURL.DropURL(DropServer server)` is perfectly fine for generating random DropURLs.

PR #317, opened 2015-04-23
Implements the send fake drop message

jasinai:
Example for future career day.


thechauffeur:
Fixes #313

src/main/java/de/qabel/core/drop/DropActor.java
```diff
@@ -320,6 +323,20 @@ public void onEvent(String event, MessageInfo info, Object... data) {
...
+	static class FakeMessage extends ModelObject {
+		public String content;
+
+		public FakeMessage(String content) {
+			this.content = content;
+		}
+	}
+
+	public static void sendFakeMessage(EventEmitter emitter, Contact c) {
+		DropMessage<? extends ModelObject> dm = new DropMessage<FakeMessage>(c.getContactOwner(), new FakeMessage("dgjshsjgh"));
```
thechauffeur: The fake message is always the same. You should randomly create this message. @L-Henke said this is possible with the noise box.

src/main/java/de/qabel/core/drop/DropActor.java
```diff
@@ -320,6 +323,20 @@ public void onEvent(String event, MessageInfo info, Object... data) {
...
+		Contact fakeContact = new Contact(c.getContactOwner(), c.getDropUrls(), new QblECKeyPair().getPub());
```
thechauffeur: The fake message will be send to the same destination as the original message. This will not help mitigating the meta data :)

thechauffeur:
It would be better to think about how to solve this problem (after ackack has been implemented completely) rather than develop something fishy.

PR #318, opened 2015-04-24
Issue298 part2 port config component to ackack

### LINKED ISSUE #298: Port Config and Contact components to ackack

f-porter:
Depends on PR #305.

Closes #298.


f-porter:
Again a time out error in travis...

L-Henke:
What is the purpose of the ```replace``` method? Wouldn't it replace the object with the same object? I would an replace method like this ```replace(oldObject, newObject)```, your ```replace(object)``` is more like an ```update(object)```.

thechauffeur:
#305 is merged now please update this PR accordingly.

f-porter:
@L-Henke, you are right. From this view the ```replace``` method is like an ```update``` method. I will change the name to prevent misinterpretation.

PR #321, opened 2015-04-27
Fixes #303

### LINKED ISSUE #303: Define how to deal with dependent method calls

thechauffeur:
Define how to deal with depended methods calls.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,6 +66,8 @@ Regarding the look and feel most -- but not all -- of this is reflected by the E
 * Try to fit your code into **80** characters but do not wrap your line of code just for the sake of wrapping it. Use **120** in your auto format settings.
 * Statements like ```if``` always with ```{ }```. Do not just write ```if (foo) bar;```
 * Our code does not need to be compatible to versions below Java 7. That means you can and should use Java 7 features and syntax.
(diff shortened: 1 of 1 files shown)


thechauffeur:
Fixes #303

L-Henke:
:+1:

jan-schreib:
The name of the method has to be ```initialize()```? IIRC @L-Henke named one method ```init()``` which was renamed to ```connect()``` eventually. Don't get me wrong, I'm fine with the longer name. I'm just wondering because we had ```init()``` in a pullrequest about 4 hours ago.

L-Henke:
E.g. means it's just an example.

jan-schreib:
Misinterpreted the 'E.g.' there. :+1:

Issue #307, opened 2015-03-30
Compile curve25519 lib with gradle or provide binaries for all supported operating systems


jan-schreib:
It would be nice to have an extra option to recompile the shared library for the jni component.

thechauffeur:
@moritz4004 ask @jan-schreib if you need help.

jan-schreib:
@moritz4004 whats the status on this?

moritz4004:
I have created a pull request. There is an Error, if you try gradle build in Qable Core.

thechauffeur:
Just as a side note. If this will not work in a timely manner we can provide binaries for the other operating systems. So this is really an enhancement.

thechauffeur:
I changed the meaning of this issue a little bit because we need to address this until a BETA release and not after that.

thechauffeur:
@hartmannaf please ask @moritz4004 what is going ob with this and tell me or update this issue (and assign yourself).

cburkert:
@moritz4004 Now that you're back, please re-tackle this issue.

jasinai:
I think, this issue is closed by Qabel/qabel-core#407, or is there more to do?

jan-schreib:
We are not able to build for the Apple operating system right now, we could however close this one and open another for just the Apple part.

audax:
And I just did exactly that.

PR #325, opened 2015-04-30
Autocompile lib Curve

moritz4004:
Pull request for Issue #307

Add Code to build.gradle for complie Curve25519 for exactly your OS, remove libcurve25519.so, because the file is no longer needed. 

The code only works if you make gradlew build in the qabel repo.
In qable core you get this Error:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/home/moritz/Dokumente/foo/qabel-core/build.gradle' line: 125

* What went wrong:
A problem occurred configuring root project 'qabel-core'.
> Exception thrown while executing model rule: org.gradle.nativeplatform.plugins.NativeComponentModelPlugin$Rules#createNativeBinaries(org.gradle.platform.base.BinaryContainer, org.gradle.api.NamedDomainObjectSet<org.gradle.nativeplatform.NativeComponentSpec>, org.gradle.language.base.internal.LanguageRegistry, org.gradle.nativeplatform.toolchain.internal.NativeToolChainRegistryInternal, org.gradle.platform.base.PlatformContainer, org.gradle.nativeplatform.BuildTypeContainer, org.gradle.nativeplatform.FlavorContainer, org.gradle.internal.service.ServiceRegistry, java.io.File)
   > Could not find property 'builder' on task ':linkLinux_i386Curve25519SharedLibrary'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```


build.gradle
```diff
@@ -27,12 +28,114 @@ dependencies {
 	compile project(':ackack')
 }
 
+task nativeHeaders {
+	String inputClass = 'de.qabel.core.crypto.Curve25519';
```
L-Henke: Gradle don't needs a ; at the line ending.

L-Henke:
Maybe it would have been better to copy the script from the original source https://github.com/adammurdoch/native-platform/blob/master/build.gradle and not the already modified version. The original script can also handle OS X and FreeBSD. I further think, that unless it is absolutely impossible the qabel-core repo should remain buildable on its own.

cburkert:
@moritz4004 You told me you're stuck here. Please describe what you tried.

BTW: Did you notice @L-Henke suggestions?

jan-schreib:
closed in favor of #407

PR #329, opened 2015-05-13
Refactored DropActor

### LINKED ISSUE #327: Refactor DropActor
The current DropActor has getter and setter for ```DropServers```, ```Identities``` and ```Contacts```. Since we have (partly) merged the Contact and Config Actors, we should remove these methods from the DropActor. Only the private ```retrieve()``` method seems to be using these members. This could then be refactored to use the default actors.

### LINKED ISSUE #336: Update methods in entity collections are not working as intended
All update methods in the entity collections like [Accounts](https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/Accounts.java#L35) are not working. They are storing the entities in HashSets and are first trying to remove the entity by calling ```remove()``` with the already changed entity. Since this cannot be found in the HashSet, the update method always adds a new entity and never replaces them.

L-Henke:
Refactored DropActor to get resources via events. ~~The ContactActorTest has some concurrency issues, because modifying contacts is done in a different thread, which cannot be joined anymore. This might be avoided with a simple sleep, but maybe there is a cleaner solution.~~ EDIT: ContactActorTest now listens to contact added and removed events to wait until events has been processed.

- [X] Get Contacts via event
- [x] Get Identities via event
- [x] Get DropServers via event
- [x] Eliminated concurrency problems with ContactActorTest

~~This is related to #327, but does not closes it in the current state.~~
~~The unregister method from https://github.com/Qabel/ackack/pull/7 is required for this PR.~~ -> merged

~~Synchronization may get further optimized.~~ -> done

~~Requires #334~~ -> merged 
- [x] Fix timeout with large persistence file
- [x] Default LocalSettings -> #343 
- [X] Define when to emit config changes -> #344 
- [x] Check synchronization in Config and ContactActor

Fixes #336 
Closes #327

### DIFF
--- a/src/main/java/de/qabel/core/config/Accounts.java
+++ b/src/main/java/de/qabel/core/config/Accounts.java
@@ -1,26 +1,32 @@
 package de.qabel.core.config;
 
-import java.util.Collections;
-import java.util.Set;
-import java.util.HashSet;
+import java.util.*;
 
 /**
  * https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Configuration#accounts
  */
 public class Accounts {
 
-	private final Set<Account> accounts = new HashSet<Account>();
+	private final Map<String, Account> accounts = new HashMap<>();
 
 	/**
 	 * Returns unmodifiable set of contained accounts
 	 * @return Set<Account>
 	 */
 	public Set<Account> getAccounts() {
-		return Collections.unmodifiableSet(this.accounts);
+		return Collections.unmodifiableSet(new HashSet<>(this.accounts.values()));
 	}
-	
+
+	/**
+	 * Adds or updated an account
+	 * @param account Account to add or update
+	 * @return True if newly added, false if updated
+	 */
 	public boolean add(Account account) {
-		return this.accounts.add(account);
+		if (this.accounts.put(account.getPersistenceID(), account) == null) {
+			return true;
+		}
+		return false;
 	}
 
 	/**
@@ -29,12 +35,10 @@ public class Accounts {
 	 * @return true if account was contained in list, false if not
 	 */
 	public boolean remove(Account account) {
-		return this.accounts.remove(account);
-	}
-
-	public boolean update(Account account) {
-		this.remove(account);
-		return this.add(account);
+		if (this.accounts.remove(account.getPersistenceID()) == null) {
+			return false;
+		}
+		return true;
 	}
 
 	@Override
@@ -61,7 +65,4 @@ public class Accounts {
 			return false;
 		return true;
 	}
-	
-	
-
 }
--- a/src/main/java/de/qabel/core/config/Accounts.java
+++ b/src/main/java/de/qabel/core/config/Accounts.java
@@ -18,11 +18,11 @@ public class Accounts {
 	}
 
 	/**
-	 * Adds or updated an account
-	 * @param account Account to add or update
+	 * Put an account
+	 * @param account Account to put
 	 * @return True if newly added, false if updated
 	 */
-	public boolean add(Account account) {
+	public boolean put(Account account) {
 		if (this.accounts.put(account.getPersistenceID(), account) == null) {
 			return true;
 		}
--- a/src/main/java/de/qabel/core/config/ConfigActor.java
+++ b/src/main/java/de/qabel/core/config/ConfigActor.java
@@ -1,8 +1,11 @@
 package de.qabel.core.config;
 
 import java.io.Serializable;
+import java.util.Date;
 import java.util.Set;
 
+import de.qabel.ackack.event.EventEmitter;
+import de.qabel.core.EventNameConstants;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -17,6 +20,9 @@ import de.qabel.ackack.Responsible;
 public class ConfigActor extends Actor {
 	private static ConfigActor defaultConfigActor = null;
 	private Settings settings;
+	private EventEmitter eventEmitter;
+	private Persistence persistence;
+
 	private static final String RETRIEVE_ACCOUNTS = "retrieveAccounts";
 	private static final String RETRIEVE_DROPSERVERS = "retrieveDropServers";
 	private static final String RETRIEVE_IDENTITIES = "retrieveIdentities";
@@ -45,15 +51,55 @@ public class ConfigActor extends Actor {
 
 	private final static Logger logger = LogManager.getLogger(ConfigActor.class.getName());
 
-	static ConfigActor getDefault() {
+	public static ConfigActor getDefault() {
 		if(defaultConfigActor == null) {
-			defaultConfigActor = new ConfigActor(new Settings());
+			defaultConfigActor = new ConfigActor(new Settings(), EventEmitter.getDefault());
 		}
 		return defaultConfigActor;
 	}
 
-	public ConfigActor(Settings settings) {
-		this.settings= settings;
+	public ConfigActor(Settings settings, EventEmitter eventEmitter) {
+		this.persistence = new SQLitePersistence();
+		this.settings = settings;
+		//TODO: DEFAULT SETTINGS?!?
+		settings.setLocalSettings(new LocalSettings(1000L, new Date()));
+		settings.setSyncedSettings(new SyncedSettings());
+		this.eventEmitter = eventEmitter;
+		loadFromPersistence();
+	}
+
+	private void loadFromPersistence() {
+		for (Object object: persistence.getEntities(Account.class)) {
+			settings.getSyncedSettings().getAccounts().add((Account) object);
+		}
+
+		for (Object object: persistence.getEntities(DropServer.class)) {
+			settings.getSyncedSettings().getDropServers().add((DropServer) object);
+		}
+
+		for (Object object: persistence.getEntities(Identity.class)) {
+			settings.getSyncedSettings().getIdentities().add((Identity) object);
+		}
+
+		for (Object object: persistence.getEntities(LocaleModuleSettings.class)) {
+			settings.getLocalSettings().getLocaleModuleSettings().add((LocaleModuleSettings) object);
+		}
+
+		for (Object object: persistence.getEntities(LocalSettings.class)) {
+			settings.setLocalSettings((LocalSettings) object);
+		}
+
+		for (Object object: persistence.getEntities(StorageServer.class)) {
+			settings.getSyncedSettings().getStorageServers().add((StorageServer) object);
+		}
+
+		for (Object object: persistence.getEntities(StorageVolume.class)) {
+			settings.getSyncedSettings().getStorageVolumes().add((StorageVolume) object);
+		}
+
+		for (Object object: persistence.getEntities(SyncedModuleSettings.class)) {
+			settings.getSyncedSettings().getSyncedModuleSettings().add((SyncedModuleSettings) object);
+		}
 	}
 
 	/**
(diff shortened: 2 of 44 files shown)


L-Henke:
Nearly all ConfigActorTests had to be redone, so you @Qabel/developer should have a look if they still test the correct things.

This PR now changes some more stuff than initially intended (like adding the persistence into the ConfigActor), but it was almost impossible to separate all the different required actions to refactor the DropActor.

I had a timeout error with the drop related tests when a large persistence file had to be read in. And we have to think about how the default LocalSettings should be set and which config changes has to be emitted.

src/main/java/de/qabel/core/config/Accounts.java
```diff
@@ -1,26 +1,32 @@
...
 	 * @return Set<Account>
 	 */
 	public Set<Account> getAccounts() {
-		return Collections.unmodifiableSet(this.accounts);
+		return Collections.unmodifiableSet(new HashSet<>(this.accounts.values()));
 	}
-	
+
+	/**
+	 * Adds or updated an account
```
jasinai: "Adds or __updates__ an account"

src/test/java/de/qabel/core/MultiPartCryptoTest.java
```diff
@@ -59,7 +64,12 @@ public void setStr(String str) {
 
     @Before
     public void setUp() throws InvalidKeyException, MalformedURLException, QblDropInvalidURL, InterruptedException {
-        emitter = new EventEmitter();
+        Persistence.setPassword(encryptionPassword);
+        contactsActorThread = new Thread(ContactsActor.getDefault());
+        contactsActorThread.start();
+		configActorThread = new Thread(ConfigActor.getDefault());
+		configActorThread.start();
```
jasinai: Mixed indentation in this file

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -84,36 +103,40 @@ public boolean removeContacts(final String... keyIdentifiers) {
...
+						this.contacts.add(contact);
+						persistence.updateOrPersistEntity(contact);
+						eventEmitter.emit(EventNameConstants.EVENT_CONTACT_ADDED, contact);
+					}
+					break;
+				case RETRIEVE_CONTACTS:
+					Contact[] contactsArray;
+					if(data.length > 0) {
+						ArrayList<Contact> contactsList = new ArrayList<>();
+						for (Object object : data) {
```
aBothe: ´synchronized´ should surround this for-loop only

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -84,36 +103,40 @@ public boolean removeContacts(final String... keyIdentifiers) {
...
-			info.response((Serializable[]) contactsArray);
-			break;
-		case REMOVE_CONTACTS:
-			for (int i = data.length-1; i>=0; i--) {
-				persistence.removeEntity(contacts.getByKeyIdentifier(data[i].toString()).getPersistenceID(), Contact.class);
-				this.contacts.remove(data[i].toString());
+		synchronized (contacts) {
+			switch (info.getType()) {
+				case WRITE_CONTACTS:
+					for (Object object : data) {
```
aBothe: Does `WRITE` mean overwrite? Or just append?

src/main/java/de/qabel/core/config/ContactsActor.java
L-Henke: The ContactActor and ConfigActor will get merged into one Actor (#335). Then the cases will look like in the ConfigActor.

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -84,36 +103,40 @@ public boolean removeContacts(final String... keyIdentifiers) {
...
+							contactsList.add(this.contacts.getByKeyIdentifier((String) object));
+						}
+						contactsArray = contactsList.toArray(new Contact[0]);
+					} else {
+						contactsArray = this.contacts.getContacts().toArray(new Contact[0]);
+					}
+					info.response((Serializable[]) contactsArray);
+					break;
+				case REMOVE_CONTACTS:
+					for (Object object: data) {
```
aBothe: Here as well: `synchronized` should only wrap this for-loop

src/main/java/de/qabel/core/config/ContactsActor.java
L-Henke: Also overwrite, but that is not introduced by this PR.

src/main/java/de/qabel/core/config/ContactsActor.java
aBothe: okay

src/main/java/de/qabel/core/config/Persistable.java
```diff
@@ -0,0 +1,32 @@
+package de.qabel.core.config;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+public abstract class Persistable implements Serializable {
```
thechauffeur: A class javadoc would be nice explaining the purpose of this class.

src/main/java/de/qabel/core/config/Persistable.java
```diff
@@ -0,0 +1,32 @@
...
+		this.persistenceID = genPersistenceID();
+	}
+
+	public String getPersistenceID() { return persistenceID.toString(); }
+
+	/**
+	 * Generated a random UUID for persistent storage. It ensures that Entities are using a unique UUID.
+	 * @return Unique UUID
+	 */
+	private UUID genPersistenceID(){
```
thechauffeur: Two things.
First a UUID should be unique by definition. If we manually try to ensure uniqueness here we would need to do this elsewhere, too.
Second the way you want to ensure uniqueness doesn't look good in a multi threaded environment.

src/main/java/de/qabel/core/config/Persistable.java
```diff
@@ -0,0 +1,32 @@
+package de.qabel.core.config;
+
+import java.io.Serializable;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+public abstract class Persistable implements Serializable {
+
+	private static final Set<UUID> uuids = new HashSet<>();
```
thechauffeur: Does this get persisted? You know static member, serializable, program restarts etc.

src/main/java/de/qabel/core/config/Persistable.java
L-Henke: A normal UUID is unique, because it uses timestamps etc. A random UUID is just as unique as every random value. But this should be unique enough, so we can just get rid of the check. Haven't thought about multi threading...

src/main/java/de/qabel/core/config/Persistable.java
```diff
@@ -0,0 +1,32 @@
...
+
+	private UUID persistenceID;
+
+	public Persistable() {
+		this.persistenceID = genPersistenceID();
+	}
+
+	public String getPersistenceID() { return persistenceID.toString(); }
```
thechauffeur: Why converting the UUID to a string instead of just returning a UUID?

L-Henke: Because I see no advantage in using the UUID class, since a UUID is just a string. Especially a random UUID doesn't profit from the few methods provides from the UUID class like ```getTimestamp()``` or ```clockSequence```. Because it actually is nothing more than a random string, which just looks like a UUID.

L-Henke: And no class outside of the Persistence class has to know that the persistence ID is a UUID. It just have to know that it is a unique string.

src/main/java/de/qabel/core/config/Persistence.java
```diff
@@ -100,19 +100,24 @@ final public boolean changePassword(char[] oldPassword, char[] newPassword) {
 
 	/**
 	 * Persists an entity
-	 * @param id ID for entity storage
 	 * @param object Entity to persist
 	 * @return Result of the operation
 	 */
-	abstract public boolean persistEntity(String id, Serializable object);
+	abstract public boolean persistEntity(Persistable object);
```
thechauffeur: Please update the javadoc (e.g. javadoc for line 19 in this class) to reflect that the objects must be `Persistable` instead of `Serializable`.

thechauffeur: If you only persist `Persistable` objects you can update the method signatures to retrieve `Persistable` objects instead of `Object`.

Sorry for commenting these things on this line but I cannot comment on lines you didn't patch -- which is the problem :D

src/main/java/de/qabel/core/config/ConfigActor.java
```diff
@@ -335,158 +381,231 @@ public boolean removeSyncedModuleSettings(final SyncedModuleSettings...syncedMod
...
-				.getSyncedModuleSettings();
-			for(int i = 0; i< data.length; i++) {
-				syncedModuleSettingsList.remove((SyncedModuleSettings) data[i]);
-			}
-			break;
-		default:
-			logger.debug("Unexpected type of MessageInfo: \"" + info.getType()
-					+ "\"");
-			break;
+		switch (info.getType()) {
```
thechauffeur: If you just changed the damn indentation please revert the commit introducing this. I don't want to review these lines.

src/main/java/de/qabel/core/config/ContactsActor.java
```diff
@@ -28,20 +35,32 @@ public static ContactsActor getDefault() {
 	private static final String RETRIEVE_CONTACTS = "retrieveContacts";
 	private static final String REMOVE_CONTACTS = "removeContacts";
 
+	/**
+	 * Creates a new ContactActor with given contacts. Uses the default EventEmitter.
+	 * @param contacts Contacts to use in ContactActor
+	 */
 	public ContactsActor (Contacts contacts) {
```
thechauffeur: I guess these constructors are public because of our tests, right?

src/main/java/de/qabel/core/config/ConfigActor.java
L-Henke: In that case, please have a closer look before ranting. All these parts have been changed. Cases are in synchronized blocks, events get emitted, for loops has been replaced with more elegant for each loops, objects are only casted once possible and objects gets persisted. And since every single case has been modified. Since diff couldn't track these changes individually anymore, I also fixed the wrong indentation. Do I look like I would include such mess in this kind of PR without a reason?

src/main/java/de/qabel/core/config/ContactsActor.java
thechauffeur: Is this constructor ever needed solely? Or will it just be called from the other constructors? If so please do not provide this constructor.

L-Henke: As fas as I know, no one is forced to use the default actors. That's why there is a public constructor.

src/main/java/de/qabel/core/config/DropServers.java
```diff
@@ -1,8 +1,6 @@
 package de.qabel.core.config;
 
-import java.util.Collections;
-import java.util.Set;
-import java.util.HashSet;
+import java.util.*;
```
thechauffeur: It doesn't really matter with these lines but please keep in mind that IDE automagic changes are not helpful for the reviewers of such big PRs...

L-Henke: I have already disabled all import magic in IntelliJ, but as soon as there are more than X imports from the same class, they get replaced with *. If you know which well hidden option disables this, I would be grateful.

src/main/java/de/qabel/core/config/EntityMap.java
```diff
@@ -1,51 +1,39 @@
...
- * EntityMaps provide funcionality to lookup an Enity based
+ * EntityMaps provide functionality to lookup an Entity based
  * on its key identifier.
  * 
  * @see Entity
  */
 abstract public class EntityMap<T extends Entity> implements Serializable {
 	private static final long serialVersionUID = -8004819460313825206L;
-	private final Map<String, T> entities = new HashMap<>();
+	private final Map<String, T> entities = Collections.synchronizedMap(new HashMap<String, T>());
```
thechauffeur: This looks cool. Will this magically make this map thread safe? If so can you need this for the `UUID` `map` in `Persistable`? Also I guess this has not been done in `Accounts` and/or `DropServers` because we use a `Set` there, right?

src/main/java/de/qabel/core/config/SyncSettingItem.java
```diff
@@ -6,7 +6,7 @@
 /**
  * Class SyncSettingItem: Class to store common setting for synced item
  */
-public class SyncSettingItem implements Serializable {
+public class SyncSettingItem extends Persistable implements Serializable {
```
thechauffeur: Why does it implement `Serializable`? It already extends `Persistable`.

src/main/java/de/qabel/core/config/EntityMap.java
L-Henke: Not magically, but semi magically. This is what the Javadoc says:
```
Returns a synchronized (thread-safe) map backed by the specified map. In order to guarantee serial access, it is critical that all access to the backing map is accomplished through the returned map.

It is imperative that the user manually synchronize on the returned map when iterating over any of its collection views:

  Map m = Collections.synchronizedMap(new HashMap());
      ...
  Set s = m.keySet();  // Needn't be in synchronized block
      ...
  synchronized (m) {  // Synchronizing on m, not s!
      Iterator i = s.iterator(); // Must be in synchronized block
      while (i.hasNext())
          foo(i.next());
  }
 ```

There is also a synchronized set method http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedSet%28java.util.Set%29, but I haven't included it yet.

File: src/main/java/de/qabel/core/config/LocalSettings.java [hunk 16]
```diff
@@ -14,7 +14,7 @@
 /**
  * https://github.com/Qabel/qabel-doc/wiki/Qabel-Client-Configuration#local-settings
  */
-public class LocalSettings implements Serializable {
+public class LocalSettings extends Persistable implements Serializable {
```
thechauffeur: Why does it implement `Serializable`? It already extends `Persistable`.

src/main/java/de/qabel/core/config/SQLitePersistence.java
```diff
@@ -152,13 +152,11 @@ boolean reEncryptMasterKey(KeyParameter oldKey, KeyParameter newKey) {
...
 	private void createTables(){
-		try {
-			String sql = "CREATE TABLE IF NOT EXISTS CONFIG " +
-					"(ID TEXT PRIMARY KEY NOT NULL," +
-					"DATA TEXT NOT NULL)";
-			Statement statement = c.createStatement();
+		String sql = "CREATE TABLE IF NOT EXISTS CONFIG " +
+				"(ID TEXT PRIMARY KEY NOT NULL," +
+				"DATA TEXT NOT NULL)";
+		try (Statement statement = c.createStatement()){
```
thechauffeur: I like this try with resource things :) Nevertheless there is a missing whitespace :P

src/main/java/de/qabel/core/config/SQLitePersistence.java
```diff
@@ -317,6 +298,16 @@ public boolean updateEntity(String id, Serializable object) {
 	}
 
 	@Override
+	public boolean updateOrPersistEntity(Persistable object) {
```
thechauffeur: Please add javadoc.

thechauffeur: Sorry, nvmd.

L-Henke: Javadoc is as always in the abstract super class:
```
	/**
	 * Updates a previously stored entity or persist a new entity
	 * @param object Entity to replace stored entity with
	 * @return Result of the operation
	 */
```

L-Henke: Too fast :)

src/main/java/de/qabel/core/config/StorageServers.java
```diff
@@ -28,14 +28,13 @@ protected StorageServer getStorageServerByUrl(String serverUrl) {
 	/**
 	 * Adds a storage server
 	 * @param storageServer StorageServer to add.
-	 * @return true if successfully added, false if already contained
+	 * @return true if newly added, false if updated
 	 */
 	public boolean add(StorageServer storageServer) {
```
thechauffeur: IIRC you have more methods of this style in this PR. Why do you want your `add` methods to also update stuff. The code is lying now.

thechauffeur: I'm not meaning that methods with this functionality are bad or useless but the name `add` is wrong. I would name it something like `addOrUpdate` or just `put` :)

L-Henke: I also just thought that put is probably better than add.

src/main/java/de/qabel/core/drop/DropActor.java
```diff
@@ -293,7 +320,7 @@ private DropResult send(DropMessage<? extends ModelObject> message, Collection<C
 					break;
 				}
 				if (dropMessage != null) {
-					for (Contact c : contacts) {
+					for (Contact c : ccc) {
```
thechauffeur: Thanks for fixing an unknown bug (the shuffled list hasn't been used until now).

L-Henke: I really was surprised by that. I was pretty sure, that it had been used.

src/main/java/de/qabel/core/drop/DropActor.java
```diff
@@ -48,15 +50,72 @@ public long getInterval() {
 	public DropActor(EventEmitter emitter) {
 		super(emitter);
 		this.emitter = emitter;
+		this.mContacts = new Contacts();
```
thechauffeur: I know why we need this and why we need to react on events for contacts etc. but someone who doesn't know this is like WTF. Please write Javadoc explaining this whole... let's call it workaround for now.

src/main/java/de/qabel/core/config/Persistable.java
thechauffeur: You are right that there is no advantage in form of provided methods by the UUID class and such. But the nomenclature has meaning. A `time` is technically just a `long` but you would return a `time` because it provides meaning.

I leave it up to you.

src/main/java/de/qabel/core/config/ConfigActor.java
thechauffeur: For example the case line didn't change and it would be much easier to have a case by case diff. Even this would help. I mean just look at the diff now. A human has to manually diff what is the same and what not. You don't need to get angry.

src/main/java/de/qabel/core/config/ContactsActor.java
thechauffeur: I've heard this about default emitters but not about a contacts actor. I mean should a module really instantiate its own e.g. contacts actor? I don't think so.

src/main/java/de/qabel/core/config/DropServers.java
thechauffeur: No I don't know that but I can tell you how I handle problems like this. I use git add -p so that I -- the developer -- can decide what chunks will be committed. Git gui is also a nice tool for that.

L-Henke: I have used git gui with chunk and line commits, but I didn't check for the changed imports.

thechauffeur:
So, @cburkert and I reviewed everything in its current (prior to the `merge master` commit) state except for the tests.

src/main/java/de/qabel/core/config/ContactsActor.java
L-Henke: Maybe. But neither the modifier of ContactsActor () nor ContactsActor (Contacts contacts) has been changed by this PR. Of course I can change it in this PR, although it might be a own issue.

src/main/java/de/qabel/core/config/Persistable.java
L-Henke: As I tried to state in the last comment, I want to use a UUID as the persistence ID, because its an easy way to create a random and highly probably unique string without implementing it by ourself, but no other class should have to know how the persistence ID has been created. So for everyone outside, it should be just a string. Maybe we want to change the way how these IDs are created later. This wouldn't affect anyone if we just provide a string, but it would if we give out this string as a UUID instance.

src/main/java/de/qabel/core/config/ContactsActor.java
thechauffeur: Yes. If it is an issue then it is another issue and doesn't need to be fixed by this PR.

src/main/java/de/qabel/core/drop/DropActor.java
L-Henke: Is that really just a workaround?

src/main/java/de/qabel/core/config/Persistence.java
L-Henke: Done

src/main/java/de/qabel/core/drop/DropActor.java
L-Henke: Done

thechauffeur:
@jasinai look at the tests. Look if they test the stuff they tested before and if they make sense. Good tests would be nice here ;)

jasinai:
Tests look good and make sense. I like the clean up of the setups in the test methods.

jasinai:
Take instead of poll is muuuch better :)

L-Henke:
qabel-desktop adapted to changes with: https://github.com/Qabel/qabel-desktop/pull/10

thechauffeur:
When qabel-helloworld-module is also adapted (i.e. still builds) to the changes introduced with this PR this PR can be merged (and the PRs needed for adaption, too).

PR #330, opened 2015-05-18
Merged ConfigActor and ContactsActor into ResourceActor

L-Henke:
Since the ConfigActor and the ContactsActor are very similar, I think it would be easier for a developer to provide the config values (like identities, drop server, etc) and the contacts by the same actor.


L-Henke:
Since the ConfigActorTest has massively changed in #329, I'll close this PR and redo this after #329 is merged.

PR #334, opened 2015-05-18
Fixed errors with unclosed resources in SQLitePersistence

L-Henke:
All resources are now auto-closed by usage of try with resources.
Edit: An exception while executing a statement could leave the database locked. Now try with resources are ensuring, that the statements and result sets are closed in every case. https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html


jasinai:
Looks good and I like it :+1:

Issue #248, opened 2015-01-15
Refactoring TypeAdapter of config classes 


f-porter:
Currently the TypeAdapters return null when the JSON format is invalid. In order to avoid this null returns they have to be changed to throw parsing exceptions.

f-porter:
@cburkert we talked about this last monday:
For properly throwing parsing exceptions I will have to add custom TypeAdapters for every config item and call the TypeAdapter for SyncSettingItem class there. Therefore the SyncSettingItem has to be - despite the documentation - serialized as nested object in the particular item. You were not that happy about using so many custom TypeAdapters, have you found another way to realize that? Otherwise I will go ahead and change the JSON format as described (..and add the necessary TypeAdapters).

cburkert:
@f-porter Unfortunately, I didn't have to investigate this so far. I'd say go ahead with your approach.

thechauffeur:
@cburkert I assigned this to you because at least you talked with @f-porter about this. Please label this (if it makes sense) and set a milestone.

cburkert:
I changed my mind about the urgency of this issue. As long as we are the only ones how create jsons, we can assume the correctness. If we need interoperability we should tackle this issue again.

PR #337, opened 2015-05-29
Deserialization validator example

f-porter:
Refers to #248.

GSON does not provide a proper functionality to detect missing items in a JSON input without using a custom TypeAdapter for each class. This PR is a proposal how that issue could be solved.
Therefore please give feedback.

EDIT: Don't merge this, it is just an example.

### DIFF
--- a/src/main/java/de/qabel/core/config/Account.java
+++ b/src/main/java/de/qabel/core/config/Account.java
@@ -9,6 +9,7 @@ public class Account extends SyncSettingItem {
 	 * Provider of the account
 	 * Field name in serialized json: "provider"
 	 */
+	@RequiredInJson
 	private String provider;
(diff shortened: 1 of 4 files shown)


aBothe:
Imho, this should be called null-check as it's not checking whether the fields are actually in the JSON but just checking for the field instances to be not null.

What if `auth` is explicitly set in JSON but still empty (not-null but `""`)?
There is more logic required for checking these fields anyway.

thechauffeur:
I assigned @cburkert because I assigned him to the issue, too.

cburkert:
I'm frightened about the impact of this solution. But I think we cannot avoid something like that.
However, this is not critical as long as we don't need interoperability. See #248.

Issue #346, opened 2015-06-09
Use own database in test environment


L-Henke:
We want to extend the ResourceActor to enable usage of a specific database. This will be used to execute all tests using persistence resources with an own database for testing purposes.

PR #351, opened 2015-06-11
Renaming/Changing URL to URI in almost every file

### LINKED ISSUE #340: Change URL to URI
We use URL in several classes, but this brings problems with it.
Comparing two urls:
- makes a DNS lookup, which can be a blocking operation
- they are equal, if they have the same IP address

I think, we should change it from URL to URI.

jasinai:
Look, if I changed every necessary variables/names/classes. :)
I didn't change DropURL to DropURI, after discussing with @L-Henke .
Closes #340


cburkert:
You forgot to update qabel-desktop. That's travis is failing (the setter in QblMain).

src/test/java/de/qabel/core/config/UriTestFactory.java
```diff
@@ -0,0 +1,27 @@
...
+	int i = 0;
+
+	@Override
+	public URI create() {
+		URI uri = null;
+		try {
+			uri = new URI("http://just.a.url.com/" + i++);
+		} catch (URISyntaxException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
cburkert: Maybe you can remove this stack trace, too. A runtime exception should be preferred.

cburkert:
I'll provide a path for qabel-desktop

src/main/java/de/qabel/core/config/StorageVolume.java
```diff
@@ -64,7 +64,7 @@ public StorageServer getStorageServer() {
 
 	public void setStorageServer(StorageServer storageServer) {
 		this.storageServer = storageServer;
-		this.storageServerUrl = storageServer.getUrl().toString();
+		this.storageServerUrl = storageServer.getUri().toString();
```
cburkert: Why didn't you change the variable name `storageServerUrl`?

src/main/java/de/qabel/core/config/StorageServer.java
```diff
@@ -19,28 +19,28 @@
...
-		this.setUrl(url);
+	public StorageServer(URI uri, String auth) {
+		this.setUri(uri);
 		this.setAuth(auth);
 	}
 	
 	/**
-	 * Returns the url of the StorageServer.
+	 * Returns the uri of the StorageServer.
 	 * @return URL
```
cburkert: This should be URL then

cburkert:
A few left urls in variable names (I don't know if we want to change them):
* [StorageServers](https://github.com/jasinai/qabel-core/blob/URLToURI/src/main/java/de/qabel/core/config/StorageServers.java#L24)
* ... (the rest is commented in the diff)

I grepped through src/main and nothing but the reported occurrences.
I guess we don't wont to change the JSON field names, so the type adapters are fine, too.

cburkert:
Together with Qabel/qabel-desktop#17, the renaming should be complete.

jasinai:
> You forgot to update qabel-desktop. That's travis is failing (the setter in QblMain).

I forgot to push it :sweat_smile: 

I pushed new commits.

src/main/java/de/qabel/core/config/StorageServers.java
```diff
@@ -31,10 +31,7 @@ protected StorageServer getStorageServerByUrl(String serverUrl) {
 	 * @return true if newly added, false if updated
 	 */
 	public boolean put(StorageServer storageServer) {
-		if (this.storageServers.put(storageServer.getUrl().toString(), storageServer) == null) {
-			return true;
-		}
-		return false;
+		return this.storageServers.put(storageServer.getUri().toString(), storageServer) == null;
```
thechauffeur: This line causes `java.lang.NullPointerException` while running the tests in many affected tests. @jasinai maybe you pushed without running the tests?

thechauffeur: Nevermind. Since our persisted objects have changed I needed to manually delete the sqlite database from our persistence component. Issues like these will be no more when #346 is solved.

thechauffeur:
A PR for `qabel-helloworld-module` is also needed. Right now the build fails because of `MalformedURLException`.

L-Henke:
This looks fine. If it is mergeable and travis passes, this can be merged.

PR #356, opened 2015-06-17
Refactored Persistence

### LINKED ISSUE #346: Use own database in test environment

L-Henke:
```Persistence``` is now a parameter of the ```ResourceActor``` constructor. This allows to externally initialize the utilized persistence class. Therefore the static ```Persistence.setPassword``` workaround is no longer required.

```ResourceActor.getDefault()``` and ```ModuleManager.getDefault()``` were removed, because they were just workaround methods and the persistence layer is not any more initializable with a default value.

The refactored persistence is utilized to used to allow a own database for each test class which is newly created in every setUp method and removed in every tearDown method. This lets every single test operate with a fresh database and avoids unwanted behaviour due to unknown values in the database.

Closes #346


jasinai:
Works with the other PRs on my machine. I have nothing to complain about this PR.

PR #362, opened 2015-07-09
Adapt EntityMap to fit the conventions of the Map interface

### LINKED ISSUE #287: Use Javadoc to document our software architecture

cburkert:
This adapts the put and remove functions to fit to the signature of Map interface of the very same name. The old functions were misleading and counter-intuitive (put returned false if already was a Entity associated to the same key identifier but replaced it anyway).

The new behaviour is how developers expect a put function to work from the Map interface.

Furthermore, I reduced to scope of the class to package private because a public scope is unnecessary for this abstract helper class.

The need for a clarification of EntityMap API has originally been expressed in https://github.com/Qabel/qabel-core/pull/305#discussion_r29244060
Closes #287

### DIFF
--- a/src/main/java/de/qabel/core/config/EntityMap.java
+++ b/src/main/java/de/qabel/core/config/EntityMap.java
@@ -22,19 +22,34 @@ abstract class EntityMap<T extends Entity> implements Serializable {
 		return Collections.unmodifiableSet(new HashSet<>(entities.values()));
 	}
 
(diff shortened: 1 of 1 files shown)


thechauffeur:
:+1:

PR #363, opened 2015-07-13
Replace Bouncy Castle with Spongy Castle

L-Henke:
Spongy Castle is a renamed version of Bouncy Castle. Since Android ships
its own and often stripped down version of Bouncy Castle, some features
are not available on Android devices. Including Bouncy Castle into the
apk doesn't help, because Android still uses the included version.

To avoid this, Spongy Castle is now used in qabel-core.


thechauffeur:
We decided not to replace Bouncy Castle with Spongy Castle because we want to support all JDKs (i.e. Oracle and Open) but Spongy Castle is not usable with the first one because it is not signed as a crypto provider.

We still need to use Spongy Castle as a crypto provider on Android because we want to ship a provider which has all the functionality we need (the stock Bouncy Castle provider differs too much from Android phone to Android phone).

We will probably address this issue in form of introducing a Spongy Castle branch to the `qabel-core` repository and auto-update it every time the `master` branch changes.

Issue #182, opened 2014-11-04
DropController.retrieve() does not work with DropServerURL w/o dropID


jasinai:
In issue qabel/qabel-doc#62 we defined, that the URL of a DropServer should not have a drop id (but the drop id is not cut off). If the DropServer URL is 'wrong', the retrieve method of the DropController does not work.
Shouldn't the DropController have at least 'my' identities with the DropURLs to retrieve the messages?

thechauffeur:
@jasinai please give an update to this issue.

PR #364, opened 2015-07-13
DropActor uses now the dropUrls of the identities instead of DropServerUrls

jasinai:
The DropserverUrls should not contain the dropIDs, so we have to
check the dropUrls of the identities.

Closes. #182


thechauffeur:
I wonder why everything is working with this short patch. It seems like we never needed the information encapsulated in DropServers. You can see this while looking at the patched test. It is strange that the test does not have to be patched more.

PR #366, opened 2015-07-14
Migrate to travis container and caching

### LINKED ISSUE #365: We should use the new Travis container-based infrastructure
Travis is notifying during builds: `This job ran on our legacy infrastructure. Please read our docs on how to upgrade`
The [upgrade guide](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade) says:
* Builds start in seconds
* Faster builds
* Caching available for open source projects
* etc.

jasinai:
Closes #365 

PS: The tests with openjdk7 and oraclejdk8 are allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ notifications:
 
 install:
  - redis-server > /dev/null 2>&1 &
- - sudo apt-get install python-pip npm nodejs
- - sudo pip install virtualenv
  - cd ..
  - git clone https://github.com/Qabel/qabel.git
  - cd qabel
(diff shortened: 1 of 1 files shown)


thechauffeur:
The Travis output for this PR is as follows.

```
Pull Request #366 Migrate to travis container and caching
Test also with openjdk7 and oraclejdk8
Because the container start so fast.
[...]
ran for 8 min 25 sec
```

We migrated because we wanted to mitigate the time needed to run a build. The `oraclejdk7` runtime is only `2min 35sec` -- this is a time I would like to see :)

@Qabel/developer what do you think?

jasinai:
All three jdks ran at the same time, so it was not really 8 min 25 sec. ;)

L-Henke:
Looks great! Faster and three different JDKs tested! :+1:

thechauffeur:
> All three jdks ran at the same time, so it was not really 8 min 25 sec. ;)

I see. :+1: then.

PR #367, opened 2015-07-15
Changed visibility of Persistence member and methods

L-Henke:
The changed visibility is required to allow Persistence implementations outside of the package.


jasinai:
:+1:

PR #369, opened 2015-07-15
Replaced LOG4J api with SLF4J api using LOG4J binding

L-Henke:
SLF4J is a logging facade which can use many different logging platforms. Here the SLF4J facade is used with the LOG4J binding to write logs into LOG4J.

Since SLF4J offers an Android binding, logging on Android is a lot easier than with LOG4J, which requires some workarounds.


jasinai:
Looks good :) And I also like the new log message in CryptoUtils.

thechauffeur:
I like the idea :+1:

PR #371, opened 2015-07-15
Added alias field to Contact

L-Henke:


jasinai:
Can you also adapt qabel-helloworld-module ?

L-Henke:
Done: https://github.com/Qabel/qabel-helloworld-module/pull/22

jasinai:
Looks good and the tests also passed.
I think, we should also update https://qabel.github.io/docs/Qabel-Client-Contact-Drop-Messages/ . The `keys` are also not up-to-date.

thechauffeur:
@jasinai yes you are right. @L-Henke should have not only done a PR for `qabel-helloworld-module` but also for the documentation. Since is has not been done and there is also no issue for that -- which is bad practice BTW -- see Qabel/qabel.github.io#97.

PR #375, opened 2015-07-16
Added exception when opening persistence with wrong password

L-Henke:
I discovered, that instantiating a persistence with a wrong encryption password isn't handled very well. This adds an exception for this case.

Requires: https://github.com/Qabel/qabel-desktop/pull/24


jasinai:
Good idea. :smiley:

PR #376, opened 2015-07-23
Replaced DropMessage payload from ModelObject to String

L-Henke:
Instead of ModelObject, Strings are now used as the payload for a
DropMessage. dropPayload is the actual payload while dropPayloadType
describes its type.

The advantage of this solution is, that the qabel-core doesn't know the
actual ModelObject type. Instead the a raw String is handed to the
receiver of a DropMessage. This allows a more loose coupling of
sender/receivers to the qabel-core. This is a little bit more
inconvenient for the sender and receiver of a DropMessage, because
serialization and deserialization isn't done by the qabel-core anymore.

It further removes the generic from DropMessages which wasn't actually used, because it was lost upon serialization. Thus deserializations required an unchecked cast to match the generic, which was pretty useless. 

Documentation changes will follow, if this proposal gets approved.


jasinai:
:+1: for me

thechauffeur:
TBH I don't know if I like removing the ModelObject. I would like to talk about this with you @L-Henke.

thechauffeur:
@L-Henke your proposal is accepted :+1: Go ahead.

L-Henke:
Merged master into PR.

PR #377, opened 2015-07-23
Removed unused Gson and GsonBuilder

L-Henke:
Both members were not used anywhere.


jasinai:
After the 3rd restart of the tests, travis is happy, too. :smiley:

PR #378, opened 2015-07-23
Fix enableAcknowledging in DropMessage

L-Henke:
* The method name had a spelling error
* == instead of equals was used for String comparison


jasinai:
I thought from time to time, that this word looks weird. :sweat_smile: The comment had the right spelling. :laughing:

PR #383, opened 2015-10-07
Log4j2 instead of Log4j 1.2

audax:
The included config file configures log4j2, but the build.gradle included log4j1.2 which ran without a configuration.


jan-schreib:
Good catch!

Issue #386, opened 2015-10-09
Travis does not have Python3.4 for qabel-drop and qabel-accounting


audax:
Travis does not allow multiple language entries, so we are stuck with either a free choice of JDKs or a free choice of Python versions.

If we do not specify `language: java`, we only have access to oraclejdk7. If we do not specify `language: python`, we only have access to Python 2.7 and Python 3.2.

Either we use Python3.2 for qabel-drop and qabel-accounting, which may be problematic because many libraries only support 3.3+, or we change the travis language to java and only test on oraclejdk7.

Other options:
* Switch to another CI, like for example Jenkins or drone.io which are much more flexible
* Switch back to the legacy infrastructure on travis which allows the use of sudo. With sudo we can grab Python3.4 from a PPA and install it directly.
* Switch back to the legacy infrastructure on travis and use docker containers to run qabel-drop,storage and accounting
* Run the tests with language:python and therefore only with oraclejdk7 until travis finally upgrades their containers to Ubuntu 14.04

jan-schreib:
I don't really haveopinion on this, but I strongly advise to stay in the cloud when switching our CI-System.
@Qabel/developer whats your opinion on this?

L-Henke:
Testing other JDKs is not that necessary, since we even allow them to fail. So only testing oraclejdk7 should be feasible and a very quick solution compared to setting up a new test infrastructure.

jan-schreib:
Java 7 is also the Android version, isn't it?

L-Henke:
Yes. Java 8 is currently not supported on Android.

PR #387, opened 2015-10-12
ec25519 native build

jan-schreib:
I was assigned in #325 since I don't have any rights in @moritz4004 fork, I started to work in my own.


jan-schreib:
This fixes the gradle issue that happened in #325, but right I can't travis command to use python3.4 in here. @audax do you have an idea on this?
On my test machines the lib gets build and all tests are passing.

audax:
As soon as we fix #386, it should work. I am currently working on it and just have to fix up stuff to make it ready for a PR.

jasinai:
@jan-schreib your merge was wrong, I think. With [this modified commit](https://github.com/jasinai/qabel-core/commit/6ff96b5c9f026cfdcedeb4736085e59f140a2dc8) it works as you can see [here](https://travis-ci.org/jasinai/qabel-core) (The getCredentials test fails, because I tested it in my repository. :wink: )

jan-schreib:
closed in favor of #407

PR #388, opened 2015-10-13
Integrate qabel-accounting

### LINKED ISSUE #386: Travis does not have Python3.4 for qabel-drop and qabel-accounting

audax:
This deletes the old storage API and implements the API for the accounting server which is needed for the new storage stuff.
The new storage stuff is not fully planned, but we certainly won't need the old API anymore.

* Remove old storage API
* Implement the API from qabel-accounting to retrieve AWS S3 credentials
* Fix travis for the python3.4 stuff
* Clean up travis config and make the output from gradle nicer

Fixes #386 and is needed for #387


jasinai:
Will the AccountingProfile know which AccountingServer it belongs to? Perhaps someone will have more than one profile on different servers.

audax:
I think that is something for later, when there may actually are multiple servers. For now, each application that is built on top of qable-core knows only one accounting server.

src/main/java/de/qabel/core/accounting/AccountingProfile.java
```diff
@@ -0,0 +1,35 @@
...
+	public String getPrefix() {
+		return prefix;
+	}
+
+	public void setPrefix(String prefix) {
+		this.prefix = prefix;
+	}
+
+	private Integer quota;
+	private String prefix;
```
jan-schreib: Maybe it's just me, but I always look at the top for variables. (And I just searched the shit out of the webgui to find those two.)

jan-schreib:
I don't have anything serious on this pullrequest. What concerns me are the null checks.
E.g. returning an emtpy list is way better than returning null. Avoids all this nasty, easy to forget checks. This is a general issue, I just noticed it in this pull request.
There is a test for ```getQuota()``` but there is no test that throws the exception. This is some kind of JVM testing, but tests if the logic and exception testing is done right. I also think its time for another system test, that tests alle the storage components.
I'm neither for or against a merge at this time, I'd like to get some feedback on the null checks and system test.

audax:
Could you point out the parts with the nasty nulls that could be avoided? I don't have a ton of Java experience, so I am very open for better ways to do stuff.

At the getQuota() stuff: Which exception should be testet? InvalidCredentials or the IOException?

jan-schreib:
For example ```Integer``` if you assign a default value to it, it will never be null. I think we should discuss this not in a pull request where I noticed it.

The ```getQuota()``` exceptions are probably best tested in a regression test like the MultiPartCryptoTest.java. Since qabel-accounting is a big thing, we should test the whole thing in addition to the tests you provided. The new testing will trigger both exceptions.

jan-schreib:
Merge. I do this now, because we need this. The returning null issue will be addressed in another pull request and/or issue.

PR #398, opened 2015-10-30
Local storage backend

### LINKED ISSUE #392: Box-Backend that uses local files instead of S3
For "uploading" and "downloading" into a configured folder. This backend will only be used to local unit testing of the basic functionality and does not need to emulate the quirks of S3.

A very simple backend with the following data and methods:

LocalReadBackend(localDirectory)
InputStream download(name)

LocalWriteBackend(localDirectory)
upload(name, InputStream payload)
delete(name)

audax:
These backends are just for prototyping and some local tests.
They are substitutes for the S3 backends.

Closes #392


src/main/java/de/qabel/core/storage/backends/LocalWriteBackend.java
```diff
@@ -0,0 +1,40 @@
...
+	public LocalWriteBackend(Path root) {
+		this.root = root;
+	}
+
+	@Override
+	public void upload(String name, InputStream inputStream) throws QblStorageException {
+		Path file = root.resolve(name);
+		try {
+			OutputStream output = Files.newOutputStream(file);
+			output.write(IOUtils.toByteArray(inputStream));
```
L-Henke: You are casting the whole input of the inputStream into a ```byte[]```, which means that the whole source (file) has to fit into the memory as a ```byte[]```. That's not what a *stream* is intended for and shouldn't be in the code. If this whole construct is just required for testing purposes this would be ok, but then it belongs into the ```src/test``` package.

src/main/java/de/qabel/core/storage/backends/LocalReadBackend.java
```diff
@@ -0,0 +1,28 @@
...
+import de.qabel.core.exceptions.QblStorageException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class LocalReadBackend implements StorageReadBackend {
+	private Path root;
+
```
L-Henke: Many unneeded newlines in this file. I it would just be in ```src/test``` I wouldn't mind...

Issue #393, opened 2015-10-29
Box-Backend that uses S3


audax:
For "uploading" and "downloading" into a configured root with given credentials for uploading and without credentials for downloading

S3ReadBackend(root)
InputStream download(name)

S3WriteBackend(root, credentials)
upload(name, InputStream payload)
delete(name)

audax:
Obsolete

PR #399, opened 2015-10-30
S3 storage backend

audax:
Implement the S3 backend (and a local backend for testing) according to #393


src/main/java/de/qabel/core/storage/backends/S3ReadBackend.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.http.impl.client.HttpClients;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public class S3ReadBackend implements StorageReadBackend {
+
+	String root;
```
L-Henke: Why is this package private?

src/main/java/de/qabel/core/storage/backends/S3ReadBackend.java
```diff
@@ -0,0 +1,58 @@
...
+			throw new QblStorageException(e);
+		}
+		HttpGet httpGet = new HttpGet(uri);
+		HttpResponse response;
+		try {
+			response = httpclient.execute(httpGet);
+		} catch (IOException e) {
+			throw new QblStorageException(e);
+		}
+		if (response.getStatusLine().getStatusCode() != 200) {
```
L-Henke: I think the status code should be included in the exception (if there is more than one possible error code). Now you would have to start up a debugger to get the status code.

L-Henke:
I generally dislike the backend interface design. I would not expect that any ```public``` upload method would upload an *unencrypted* object! This would be too misleading for me. In my opinion only an upload method which encrypts the object would be required outside of the storage/backend package.

jan-schreib:
@L-Henke was the backend design dislike only based on the public methods? That has been changed and I don't see any issues with it. The backend should be easy to change (read as: switch with another storage provider) and so far it is.

L-Henke:
Yes, I was complaining because the *unsecure* upload and download methods are public and we wanted to have all public accessible methods secure.

jan-schreib:
Ok, its package private now and I give a :+1: .

audax:
The code changed by now, so I'll just create a new pull request with the other changes.

PR #415, opened 2015-11-19
Spongycastle integration

roeslpa:
Removed all bouncycastle dependencies by using spongycastle. The spongycastle API is used directly, not via jce provider. Solves issue #413 .

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,8 +19,7 @@ repositories {
 dependencies {
 	testCompile group: 'junit', name: 'junit', version: '4.+'
 	testCompile group: 'org.meanbean', name: 'meanbean', version: '2.+'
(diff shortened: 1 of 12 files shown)


audax:
You changed the indentation from tabs to spaces…

roeslpa:
I changed it back, but some other format changes were automatically done by my IntelliJ, I think we use the eclipse format, right?

audax:
The more important policy is to not commit any formatting-only changes if not absolutely needed.
I don't want to be pedantic though, so it's fine by me this time.

ack

jan-schreib:
Travis is failing. You branch is not up to date.

roeslpa:
@jan-schreib both problems are solved.

PR #418, opened 2015-11-24
Create additional contsructors for QblException

### LINKED ISSUE #417: Create QblException constructor with Throwable
Instead of just allowing a string message, we should also have constructors that take an exception which caused the QblException.

audax:
This allows us to propagate the original cause for the exception.

Resolves #417


jasinai:
ack.

Issue #391, opened 2015-10-29
Create and list prefixes


audax:
Implement this interface: https://github.com/Qabel/qabel-accounting/issues/23

See #388

thechauffeur:
@audax IIRC you want to references useful stuff for @roeslpa here.

audax:
See the PR at #388 for how the accounting interface is implemented. @roeslpa

roeslpa:
Is there a specific reason why quota in AccountingProfile is an Integer and not an int?

audax:
No, no specific reason and you may change it.

audax:
The PR for the server side is merged: https://github.com/Qabel/qabel-accounting/pull/24

PR #420, opened 2015-11-30
Multiple prefixes in core

roeslpa:
Solves #391

### DIFF
--- a/src/main/java/de/qabel/core/accounting/AccountingHTTP.java
+++ b/src/main/java/de/qabel/core/accounting/AccountingHTTP.java
@@ -140,12 +140,7 @@ public class AccountingHTTP {
 				throw new IOException("No answer from login");
 			}
 			String responseString = EntityUtils.toString(entity);
-			if(responseString.length() > 6) {
-				prefixes = new ArrayList<>(Arrays.asList(
-						responseString.subSequence(2, responseString.length() - 2).toString().split("\", \"")));
-			} else {
-				prefixes = new ArrayList<>();
-			}
+			prefixes = new ArrayList<>(Arrays.asList(gson.fromJson(responseString, String[].class)));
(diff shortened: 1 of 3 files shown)


src/main/java/de/qabel/core/accounting/AccountingHTTP.java
```diff
@@ -120,6 +121,35 @@ public void updateProfile() throws IOException, QblInvalidCredentials {
...
+		HttpGet httpGet = new HttpGet(uri);
+		httpGet.addHeader("Authorization", "Token " + server.getAuthToken());
+		try (CloseableHttpResponse response = httpclient.execute(httpGet)) {
+			HttpEntity entity = response.getEntity();
+			if (entity == null) {
+				throw new IOException("No answer from login");
+			}
+			String responseString = EntityUtils.toString(entity);
+			if(responseString.length() > 6) {
+				prefixes = new ArrayList<>(Arrays.asList(
```
audax: Why don't you parse the JSON with the gson library?

src/test/java/de/qabel/core/accounting/AccountingHTTPTest.java
```diff
@@ -75,7 +75,7 @@ public void testGetCredentials() throws IOException, QblInvalidCredentials {
 
 	@Test
 	public void testGetPrefix() throws IOException, QblInvalidCredentials {
-		assertNotNull(accountingHTTP.getPrefix());
+		assertNotNull(accountingHTTP.getPrefixes());
```
audax: You should also check that the list is not empty to fully substitute the old test.

roeslpa:
Gson is now used, test is adapted and a method to create prefixes is implemented since the user had no prefix which could have been requested.

audax:
ack

Issue #421, opened 2015-12-04
Trying to decrypt a broken ciphertext should fail with an error


audax:
See:
https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/crypto/CryptoUtils.java#L209

If I try to decrypt a file that could not be decrypted (because the ciphertext was invalid for example), the decrypt method still returns true and doesn't raise an exception.

I would like to know if the decryption failed.

roeslpa:
I obviously crashed the file history of CryptoUtils in this commit https://github.com/Qabel/qabel-core/commit/688bbd4160098bcaa180ec7ed5c35ada2257120f where the error handling was already implemented.
I could apply the changes to solve this issue on top the current state or reset the changes to the state before this commit and apply the spongycastle changes again. I would prefer the second way. Any different thoughts?

audax:
Even here the error handling non existent:
https://github.com/Qabel/qabel-core/commit/963d71baaf1fc471dae4f17e39ca38a729cb037b#diff-a4f8939461a124b00f0c9b72f880dabbR804

Your commit did not change the error handling, this bug existed way before. It was a decision to ignore these errors by not writing to the file, and it was a decision that was wrong in hindsight.

A way to fix this is to return false when that exception occurs and then fix all the other code that depends on ignoring this error.

roeslpa:
After this https://github.com/Qabel/qabel-core/commit/d2ea89b3a8ef6902217dc850e442c1e23b56605d commit there was false returned when catching an exception which was unintentionally removed by my commit https://github.com/Qabel/qabel-core/commit/d2ea89b3a8ef6902217dc850e442c1e23b56605d .

audax:
Well, that makes it easier to fix. Just add a regression test for this issue and return false after catching a non-recoverable exception.

roeslpa:
Automatically closed by #422 ;)

PR #422, opened 2015-12-07
Decryption fail response

roeslpa:
Recovered false returning when decryption fails. Test to proof the correct response was added. Solves #421


roeslpa:
(it also recoveres the format of the former CryptoUtils to this https://github.com/Qabel/qabel-core/commit/58b3c16d321d72b32fa807195609aadc3f926d79 commit)

audax:
ack

PR #423, opened 2015-12-07
Changes the port from 6000 to 5000

jasinai:
The new DropServer does not like port 6000, so we use the flask standard port 5000.
Needed for qabel/qabel-drop#26

### DIFF
--- a/src/test/java/de/qabel/core/http/DropHTTPTest.java
+++ b/src/test/java/de/qabel/core/http/DropHTTPTest.java
@@ -24,18 +24,18 @@ public class DropHTTPTest {
 	public void setUp() {
 		try {
 			workingUri = new URI(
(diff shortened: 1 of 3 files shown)


src/test/java/de/qabel/core/http/DropHTTPTest.java
```diff
@@ -237,7 +237,7 @@ public void shouldContainNoNewMessagesSinceDate() {
 		assertFalse(result.isOk());
 	}
 
-	// HEAD 404 + SINCE
+	// HEAD 204 + SINCE
```
audax: huh?

jasinai: Some [lines below](https://github.com/Qabel/qabel-core/pull/423/files#diff-8d40acce88dbfe9ea088089f48675157L249) `204` is expected and not `404` and the spec also expects a `204`

jan-schreib:
:+1:

PR #426, opened 2016-01-07
fixed interface of Persistence

julianseeger:
Most methods were protected, thus couldn't be accessed from outside, now they are public.
Additionally, `getEntity` and `getEntities` returned instances of Persistence only even though the exact Class (subclass of Persistence) is known. This fixes those signatures to return the correct type.

And both commits should reference Qabel/qabel-desktop#53 , I just mixed up the issue numbers.


julianseeger:
please review @jan-schreib

jan-schreib:
Not much to it, I assume this was discussed. :+1:

julianseeger:
Nope, but in order to use it, we need public access ;)
Thx anyways

PR #427, opened 2016-01-08
Allow PBKDF2 round for DB encryption to be set externally

L-Henke:


jan-schreib:
You didn't just allow it, you are now forcing the user to set it. Wouldn't it be better to allow setting it ***and*** providing a fixed value in case the user has no idea?

L-Henke:
The user would be the developer. I think if this developer want to leave this decision to the user, he should provide a default value. 

But maybe we should add a recommendation into the javadoc of this method, so that a developer not knowing about PBKDF2 gets a reasonable value from us.

jan-schreib:
Sorry, for my bad wording. I mean the developer. Give him a reasonable value (2nd constructor that uses the value) and document that. But thats just a design decision, as long as its gernerally documented its fine with me.

L-Henke:
Good call. Done.

julianseeger:
You missed the 2nd constructor in SQLitePersistence.
It has to overwrite both constructors to make it usable

PR #431, opened 2016-01-08
implemented resetPassword for AccountingHTTP Qabel/qabel-desktop#55

julianseeger:
Client implementation of /api/v0/auth/password/reset/


jan-schreib:
There is no check if the email is valid. That checks are also missing in the Account classes. Maybe we need to look further into this.

julianseeger:
for the password reset, a validation is done server-side

jan-schreib:
Thats one request (and answer) per invalid mail. I think that could and should be avoided. The server should check nonetheless. But maybe I'm thinking to much about saving traffic right now.

audax:
The first check should happen in the desktop and android client, not in the qabel-core. In the GUI we can enforce a properly formatted email address for which both GUIs we use have decent tools available for.

Checking emails without sending an email is futile anyway.

ack

PR #433, opened 2016-01-08
added timeout for instable MultiPartCryptoTests (failing is better than stalling)

julianseeger:


audax:
Something we should have done much earlier.

Ack.

PR #434, opened 2016-01-08
extended identity and contact by email and phone

julianseeger:
according to Qabel/qabel.github.io#128


audax:
This should be fine for the moment. In the future we may want to move that to a new class for contact information so that we can reference that from an Identity and from a Contact. That would be preferable if we start to support more than just an email and a phone number.

ack

PR #440, opened 2016-01-13
Added OSX compatibility to jni tasks

thechauffeur:


thechauffeur:
Thx for the reference @jan-schreib.

PR #443, opened 2016-01-18
Extracted EncryptedPersistence from Persistence

### LINKED ISSUE #425: Allow the Persistence key to be supplied directly or to disable the encryption
~~Instead of deriving the key from a password, it should be possible to supply the key directly. This can be used on Android with a secure KeyStore.~~
In the case that there is no secure way to permanently store the key, it should be possible to disable the encryption of the persistence layer.

See https://github.com/Qabel/qabel-android/issues/131

Edited by L-Henke

L-Henke:
This allows to either implement an encrypted persistence storage or an
unencrypted persistence storage.

Closes #425

### DIFF
--- a/src/main/java/de/qabel/core/config/EncryptedPersistence.java
+++ b/src/main/java/de/qabel/core/config/EncryptedPersistence.java
@@ -1,0 +1,180 @@
+package de.qabel.core.config;
+
+import de.qabel.core.crypto.CryptoUtils;
+import de.qabel.core.exceptions.QblInvalidEncryptionKeyException;
+import org.spongycastle.crypto.InvalidCipherTextException;
+import org.spongycastle.crypto.params.KeyParameter;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Logger;
+
+import javax.crypto.SecretKey;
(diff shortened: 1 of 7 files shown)


src/main/java/de/qabel/core/config/EncryptedPersistence.java
```diff
@@ -0,0 +1,180 @@
...
+ * Persistence defines methods to store encrypted entities into a database.
+ * Entities has to be Persistable.
+ * Serialized data is encrypted with AES256 GCM, a nonce provided to the serialize/deserialize
+ * methods and a key derived from the provides password with PBKDF2 and a salt.
+ *
+ */
+public abstract class EncryptedPersistence<T> extends Persistence<T> {
+	private final static Logger logger = LoggerFactory.getLogger(EncryptedPersistence.class.getName());
+	private static final String SECRET_KEY_ALGORITHM = "PBKDF2WithHmacSHA1";
+	//TODO: Find a reasonable default value
```
jasinai: Should we open an issue for this to-do?

jasinai:
Looks good to me.

PR #444, opened 2016-01-21
Change visibility of registerSender in DropMessage to public

L-Henke:
Required to use this method from qabel-android for receiving DropMessages without the DropActor.


julianseeger:
Currently, it seems to be consistent with the constructors.
The public constructor allows direct injection of a sender. Do you have a use case example for this? ( I couldn't find related code in Qabel/qabel-android#197 )

L-Henke:
The use case is already mentioned, and the related code has been merged. https://github.com/Qabel/qabel-android/blob/master/qabelbox/src/main/java/de/qabel/qabelbox/services/LocalQabelService.java#L278

This is not a question of consistency, the requirements has changed and this registerSender workaround has to be public now. This method simply exists to avoid the limitations with deserialization by GSON.

PR #447, opened 2016-02-03
Added list of prefixes to Identity

L-Henke:
Added list of prefixes to Identity as discussed with @audax.

Issue #445, opened 2016-01-25
include prefix in exported identity


jmt85:

julianseeger:
If an identity does not include prefixes associated to box accounts, a user cannot use the same box account with the same identity on two different devices because with a new prefix, he will see other files

audax:
this is resolved, isn't it?

julianseeger:
probably yes, I didn't explicitly check the export

julianseeger:
example identity (auto-converted by gson):

```JSON
{
    "alias": "alias",
    "created": 1455145236899,
    "deleted": 0,
    "dropUrls": [
        {
            "uri": "https://inbox.qabel.de/123456789012345678901234567890123456789012c"
        }
    ],
    "email": null,
    "id": 0,
    "keys": {
        "private_key": "8c73c13069d1bf43e22d6fb2e4f3c1e765a0d15ce5abc53ef9108998168413d2"
    },
    "persistenceID": "60b7bbf8-e872-49f8-bd58-89f7026ff9b7",
    "phone": null,
    "prefixes": [
        "testprefix"
    ],
    "updated": 0
}
```

PR #448, opened 2016-02-04
Added prefixes to hashCode

roeslpa:
Resolves issue in #447

### DIFF
--- a/src/main/java/de/qabel/core/config/Identity.java
+++ b/src/main/java/de/qabel/core/config/Identity.java
@@ -112,6 +112,7 @@ public class Identity extends Entity {
 		result = super.hashCode();
 		result = prime * result + ((alias == null) ? 0 : alias.hashCode());
 		result = prime * result + ((primaryKeyPair == null) ? 0 : primaryKeyPair.hashCode());
+		result = prime * result + ((prefixes == null) ? 0 : prefixes.hashCode());
 		return result;
 	}
 
--- a/src/main/java/de/qabel/core/config/Identity.java
+++ b/src/main/java/de/qabel/core/config/Identity.java
@@ -31,8 +31,21 @@ public class Identity extends Entity {
 		prefixes = new ArrayList<>();
 	}
 
+	/**
+	 * Returns the list of prefixes of the identity
+	 * @return prefixes
+	 */
 	public List<String> getPrefixes() {
 		return prefixes;
+	}
+
+	/**
+	 * Sets the list of prefixes of the identity.
+	 * The list of prefixes is mutable, thus has no influence on the hashCode evaluation.
+	 * @param prefixes the prefixes of the identity
+	 */
+	public void setPrefixes(List<String> prefixes) {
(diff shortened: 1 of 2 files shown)


audax:
@julianseeger please confirm the fix

julianseeger:
It probably fixes the test (I'll check that later) but from an ideological point of view, after adding a prefix to an identity it is still the same identity just with some additional meta information isn't it?
When adding prefixes to an identity changes it's hash, you can't rely on an identity in identity or hash based structures like HashMaps etc. even though most information is immutable (like the private key). Maybe it's just personal opinion but I don't think so.
Other informations like email and phone don't change the hashCode and maybe the alias shouldn't chage it, too.

roeslpa:
From my point of view either all information of the identity should affect the hash or only the key pair. The glossary says key pair and nickname=alias define an identity. I cannot see a reason for attaching information to the identity directly but pretending as if they are not part of it. 
As far as I know originally contact information should not be part of the identity but "module-data". Since mail and phone number are not essential for our current goal and prefixes do not affect the identity, I would change the test instead of the hash calculation. And I would propose to outsource all mutable data to another class because there should be a way to compare attached data as well..

julianseeger:
The quick fix would probably be to extend the ConfigHashCodeTest to `ignoreProperty('prefixes')` but as we are already that deep in the design, the long version:

> From my point of view either all information of the identity should affect the hash or only the key pair. The glossary says key pair and nickname=alias define an identity.

Ack. Even though the second part suggests that the hash should be build from keypair AND alias, the latter is mutable afaik so should not be part of the hash imho.

> As far as I know originally contact information should not be part of the identity but "module-data". Since mail and phone number are not essential for our current goal and prefixes do not affect the identity, I would change the test instead of the hash calculation. And I would propose to outsource all mutable data to another class because there should be a way to compare attached data as well..

Is that neccessary? If all immutable data / properties affect the hashcode and all mutable data affect the equals method (in addition to the immutable data / hash) then all required conditions hold:
 * hashCode is immutable
 * if `a.hashCode() == b.hashCode()` now, and neither a nor b are modified then `a.hashCode() == b.hashCode()` in the future
 * if a.equals(b) then a.hashCode() == b.hashCode()
 * of course, both are reflexive, symmetric, transitive and consistent
 * as a result, `a.equals(b)` is `a way to compare attached data as well` (quote from you)

So if I didn't miss something, this could be a way to achieve everything without extracting mutable data to another class. (compare attached/mutable data, have working `equals` and `hashCode` implementations and use the class for hashing because hashes are immutable.

[SO answer to equals/hashCode](http://stackoverflow.com/questions/27581/what-issues-should-be-considered-when-overriding-equals-and-hashcode-in-java)

roeslpa:
Ack. I will apply this to the code.

roeslpa:
done.

julianseeger:
#445

julianseeger:
looks good and works for me

PR #450, opened 2016-02-09
PoW calculation

roeslpa:
This implements the calculation of a proof of work according to the spec in https://github.com/Qabel/qabel.github.io/pull/130

### DIFF
--- a/src/main/java/de/qabel/core/drop/ProofOfWork.java
+++ b/src/main/java/de/qabel/core/drop/ProofOfWork.java
@@ -3,7 +3,6 @@ package de.qabel.core.drop;
 import de.qabel.core.crypto.CryptoUtils;
 import org.spongycastle.crypto.digests.SHA256Digest;
 import org.spongycastle.util.encoders.Base64;
-import org.spongycastle.util.encoders.Base64Encoder;
 
 import java.nio.ByteBuffer;
 import java.util.Calendar;
@@ -18,7 +17,6 @@ public class ProofOfWork {
 	private byte[] IVserver;
 	private byte[] IVclient;
 	private long time;
-	private byte[] timeBytes;
 	private byte[] messageHash;
 	private long counter;
 
@@ -48,14 +46,14 @@ public class ProofOfWork {
 		this.messageHash = messageHash;
 
 		time = calendar.getTimeInMillis() / 1000L;
-		timeBytes = ByteBuffer.allocate(Long.SIZE / Byte.SIZE).putLong(time).array();
+		byte[] timeBytes = ByteBuffer.allocate(Long.SIZE / Byte.SIZE).putLong(time).array();
 		IVclient = cUtils.getRandomBytes(16);
 
 		byte[] fix = new byte[IVserver.length + IVclient.length + timeBytes.length + messageHash.length];
 		System.arraycopy(IVserver, 0, fix, 0, IVserver.length);
 		System.arraycopy(IVclient, 0, fix, IVserver.length, IVclient.length);
 		System.arraycopy(timeBytes, 0, fix, IVserver.length+IVclient.length, timeBytes.length);
-		System.arraycopy(messageHash, 0, fix, IVserver.length+IVclient.length+timeBytes.length, messageHash.length);
+		System.arraycopy(messageHash, 0, fix, IVserver.length+IVclient.length+ timeBytes.length, messageHash.length);
 
 		//Find counter which fulfills pattern
 		long i = 0;
--- a/src/main/java/de/qabel/core/drop/ProofOfWork.java
+++ b/src/main/java/de/qabel/core/drop/ProofOfWork.java
@@ -9,67 +9,56 @@ import java.util.Calendar;
 import java.util.TimeZone;
 
 public class ProofOfWork {
-	private SHA256Digest digest;
-	private Calendar calendar;
-	private CryptoUtils cUtils;
-
-	private int X;
-	private byte[] IVserver;
-	private byte[] IVclient;
+	/**
+	 * For further description of the scheme see http://qabel.github.io/docs/Qabel-Protocol-ProofOfWork/
+	 */
+	private int leadingZeros;
+	private byte[] initVectorServer;
+	private byte[] initVectorClient;
 	private long time;
 	private byte[] messageHash;
 	private long counter;
+	private static int longLength = Long.SIZE / Byte.SIZE;
+	private static int hashLength = 256/8; //SHA-256
 
 	private byte[] pow;
 	/**
 	 * Initializes PoW
 	 */
 	public ProofOfWork() {
-		calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-		cUtils = new CryptoUtils();
-		digest = new SHA256Digest();
-
 		//byte array for hash result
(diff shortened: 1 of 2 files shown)


src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+import java.nio.ByteBuffer;
+import java.util.Calendar;
+import java.util.TimeZone;
+
+public class ProofOfWork {
+	private SHA256Digest digest;
+	private Calendar calendar;
+	private CryptoUtils cUtils;
+
+	private int X;
```
audax: Just call it "leadingZeros" or something like that.

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+import org.spongycastle.util.encoders.Base64;
+
+import java.nio.ByteBuffer;
+import java.util.Calendar;
+import java.util.TimeZone;
+
+public class ProofOfWork {
+	private SHA256Digest digest;
+	private Calendar calendar;
+	private CryptoUtils cUtils;
```
audax: Please rename it to cryptoUtils

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+
+	private int X;
+	private byte[] IVserver;
+	private byte[] IVclient;
```
audax: Again: coding style. What about... "initVectorServer" and "initVectorClient" together with a short description about what they are and why they are?

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+	 * @return byte[][]: byte[0]=plain parameters byte[1]=PoW hash
+	 */
+	public byte[][] calculate(int X, byte[] IVserver, byte[] messageHash) {
+		this.X = X;
+		this.IVserver = IVserver;
+		this.messageHash = messageHash;
+
+		//time in seconds since epoch UTC
+		time = calendar.getTimeInMillis() / 1000L;
+		byte[] timeBytes = ByteBuffer.allocate(Long.SIZE / Byte.SIZE).putLong(time).array();
```
audax: Please put `Long.Size / Byte.Size` in a constant

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+		long i = 0;
+		updatePow(fix, i);
+		while(!enoughZeros(pow)) {
+			i++;
+			updatePow(fix, i);
+		}
+		counter = i;
+
+		//Byte array which contains plain pow text and pow hash
+		byte[][] result = new byte[2][Math.max(fix.length+Long.SIZE/Byte.SIZE,256/8)];
```
audax: 256/8 is a magic number

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+		//time in seconds since epoch UTC
+		time = calendar.getTimeInMillis() / 1000L;
+		byte[] timeBytes = ByteBuffer.allocate(Long.SIZE / Byte.SIZE).putLong(time).array();
+		IVclient = cUtils.getRandomBytes(16);
+
+		byte[] fix = new byte[IVserver.length + IVclient.length + timeBytes.length + messageHash.length];
+		System.arraycopy(IVserver, 0, fix, 0, IVserver.length);
+		System.arraycopy(IVclient, 0, fix, IVserver.length, IVclient.length);
+		System.arraycopy(timeBytes, 0, fix, IVserver.length+IVclient.length, timeBytes.length);
+		System.arraycopy(messageHash, 0, fix, IVserver.length+IVclient.length+ timeBytes.length, messageHash.length);
```
audax: This could be much cleaner by holding the start index and the offset index in a separate integer. You could even move the whole block into another method then, just to hide the it from the eyes of the reader, since it is just boilerplate.

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+import de.qabel.core.crypto.CryptoUtils;
+import org.spongycastle.crypto.digests.SHA256Digest;
+import org.spongycastle.util.encoders.Base64;
+
+import java.nio.ByteBuffer;
+import java.util.Calendar;
+import java.util.TimeZone;
+
+public class ProofOfWork {
+	private SHA256Digest digest;
```
audax: The digest is only used in one method, including a reset. Is it worth holding it in a member or could you just keep it local to the method?

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,114 @@
...
+	private Calendar calendar;
```
audax: It's only used in the constructor, no need for the calendar to be a member of the class.

audax:
As a general recommendation: I would do the calculation in a static method and then return a plain object containing the results. This nicely separates the dynamic parts of the ProofOfWork from the static ones.

src/test/java/de/qabel/core/drop/ProofOfWorkTest.java
```diff
@@ -0,0 +1,30 @@
+package de.qabel.core.drop;
+
+import org.junit.Test;
+import org.spongycastle.crypto.digests.SHA256Digest;
+import org.spongycastle.util.encoders.Hex;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by paul on 25.01.16.
```
julianseeger: please reconfigure your IDE

julianseeger:
> As a general recommendation: I would do the calculation in a static method and then return a plain object containing the results. This nicely separates the dynamic parts of the ProofOfWork from the static ones.

Ack. Please do not create stateful classes with an interface that suggests statelessness. This leads to hard to reproduce errors when two concurrent threads use the same instance assuming statelessness.

src/main/java/de/qabel/core/drop/ProofOfWork.java
```diff
@@ -0,0 +1,144 @@
...
+		CryptoUtils cryptoUtils = new CryptoUtils();
+		Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+		this.leadingZeros = leadingZeros;
+		this.initVectorServer = initVectorServer;
+		this.messageHash = messageHash;
+
+		//time in seconds since epoch UTC
+		time = calendar.getTimeInMillis() / 1000L;
+		byte[] timeBytes = ByteBuffer.allocate(longLength).putLong(time).array();
+		initVectorClient = cryptoUtils.getRandomBytes(16);
```
audax: `private static byte[] toByteArray(long number) { return ByteBuffer.allocate(longLength).putLong(number).array(); }`
could remove quite a bit of duplication.

audax:
We will refactor the interface later when we actually start using the code.

PR #451, opened 2016-02-09
allow concurrent cryption on CryptoUtils

julianseeger:
Currently, when the same instance of cryptoUtils is used for multiple de- or encryptions, those need to be sequential. Concurrent calls use the same `GCMBlockCipher` instance and at least one of them will fail.

cause for Qabel/qabel-desktop#125

### DIFF
--- a/src/main/java/de/qabel/core/crypto/CryptoUtils.java
+++ b/src/main/java/de/qabel/core/crypto/CryptoUtils.java
@@ -40,13 +40,11 @@ public class CryptoUtils {
 			.getName());
 
 	private SecureRandom secRandom;
-	private GCMBlockCipher gcmCipher;
 	private Mac hmac;
 	private CipherKeyGenerator keyGenerator;
 
 	public CryptoUtils() {
 		secRandom = new SecureRandom();
-		gcmCipher = new GCMBlockCipher(new AESEngine());
 		hmac = new HMac(new SHA512Digest());
 
 		//New key generator needs random for initialization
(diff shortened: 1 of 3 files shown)


audax:
That was probably ugly to debug.

julianseeger:
indeed, I was on the hunt since monday and bugs seemed to appear everywhere and vanish minutes later :-X

L-Henke:
I honestly thought that is was documented, that the CryptoUtils should NEVER be used in different threads. But this either has gone missing, or it was never documented, because all original qabel-core developers knew about this. It was actually a conscious decision we made in a meeting to not ensure any thread safety in CryptoUtils to minimize performance impacts. Further if not documented otherwise a class must always be considered not thread safe.

A minor issue, you are using if and while without brackets in your test methods. That conflicts with our contributing guidelines. Since it is just in a test is might be ignored.

julianseeger:
Maybe we can learn a lession from this:
 * low-level documentation like this that is not inside the code just doesn't exist (because you won't search or find it anyways)
 * even if it exists in the code, it is probably always outdated (self-explanatory class design over documentation)
 * thread-safety has little to do with performance impacts most of the time (on the other hand, if the using class is concurrent and wants to use the CryptoUtils, it has a performance impact by externally ensuring thread safety [mutexes or new instances all the time])

Like passing in at least the Key to the constructor shows the developer, that this class has some sort of relevant state that might impact reusability or concurrent usability. A default constructor is a bad choice imho.

L-Henke:
Creating the block cipher object is a pretty expensive operation. A quick and dirty test shows that recreating the block cipher object is about 10 times slower than just reinitializing it. But it is probably fine anyway, because the actual decryption is still by far the most expensive operation.

But if we are going to make the class thread safe, we should also make the HMAC operations thread safe.

julianseeger:
> Creating the block cipher object is a pretty expensive operation.

Don't know how good mobile devices handle that, but on desktop hardware we are talking about microseconds, probably insignificant compared to network latency.

> But if we are going to make the class thread safe, we should also make the HMAC operations thread safe.

will add a seconds PR to fix this

PR #452, opened 2016-02-10
#384 pushed owning identity from contact to contacts

### LINKED ISSUE #384: Can't add the same Contact twice (e.g. for different Identities)
Since the ```EntityMap``` stores entities in a ```HashMap``` with the ```KeyIdentifier``` as the key and the ```ResourceActor``` uses one ```EntityMap<Contact>``` to store entities, it is impossible to add two entities with the same ```KeyIdentifier```. This would be needed if two or more ```Identities``` would try to add the same ```Contact```.

See: https://github.com/Qabel/qabel-core/blob/master/src/main/java/de/qabel/core/config/EntityMap.java#L32

julianseeger:
+extended SyncedSettings.contacts from Contacts to Set<Contacts> to account for multiple identities
+removed AckAck
+removed Actors and Modules
+fixed hashCode of Contact (contained mutable alias)

This is actually a huge BC break but as neither qabel-android, nor qabel-desktop uses `ackack` the Actors shouldn't be needed anymore. As I couldn't find Modules in the spec, I removed them, too. And all Contacts-related changes are neccessary to fix #384 so they have to be added anyways.

Best way seems to be using `SyncedSettings::getContacts(Identity owner)` or just use a `Set` of Contacts who know their own `Identity`. As two Contact's with the same key are identical by hash now and don't know their owner, it should be possible to have two `Contacts` (lists) containing the same `Contact` instance.

Replacing the key-binding in `EntityMap` with some sort of `HashSet` sould have no effect because two Contacts with the same key have the same hash now.


julianseeger:
I think the new exported structure of the exported json from the SyncedSettingsTypeAdapter doesn't violate the spec. The `Client Data` from the spec doesn't include contacts anyways and the new structure is:
```JSON
{
    "accounts": [],
    "contacts": [
        {
            "contacts": [
                {
                    contact
                }
            ],
            "owner": "cbbfd5a747fc1f25028d6144c1a6fe2520176a27e4040360d43c06cd9b763763"
        }
    ],
    "drop_servers": [],
    "identities": []
}
```
(the outer) contacts is a list/set of objects containing the `contacts` element from the spec plus the key id of the owning identity. The inner `contacts` element still fits the spec and the other stuff is not specced at all.

audax:
2,393 deletions. I like this.

I'll take a look at this today.

audax:
```
diff --git a/src/main/java/de/qabel/core/config/Contacts.java b/src/main/java/de/qabel/core/config/Contacts.java
index 50e1051..99a5957 100644
--- a/src/main/java/de/qabel/core/config/Contacts.java
+++ b/src/main/java/de/qabel/core/config/Contacts.java
@@ -12,6 +12,7 @@ public class Contacts extends EntityMap<Contact> {
 	private Identity identity;
 
 	public Contacts(Identity identity) {
+		super();
 		this.identity = identity;
 	}
 
diff --git a/src/main/java/de/qabel/core/config/EntityMap.java b/src/main/java/de/qabel/core/config/EntityMap.java
index 810d2bc..a637d0d 100644
--- a/src/main/java/de/qabel/core/config/EntityMap.java
+++ b/src/main/java/de/qabel/core/config/EntityMap.java
@@ -9,11 +9,15 @@ import java.util.*;
  * 
  * @see Entity
  */
-abstract class EntityMap<T extends Entity> implements Serializable {
+abstract class EntityMap<T extends Entity> extends Persistable {
 	private static final long serialVersionUID = -4541440187172822588L;
 
 	private final Map<String, T> entities = Collections.synchronizedMap(new HashMap<String, T>());
 
+	public EntityMap() {
+		super();
+	}
+
 	/**
 	 * Returns unmodifiable set of contained contacts
 	 * 
diff --git a/src/main/java/de/qabel/core/config/Identities.java b/src/main/java/de/qabel/core/config/Identities.java
index 097ecfd..0a3a3b9 100644
--- a/src/main/java/de/qabel/core/config/Identities.java
+++ b/src/main/java/de/qabel/core/config/Identities.java
@@ -9,6 +9,10 @@ import java.util.Set;
 public class Identities extends EntityMap<Identity> {
 	private static final long serialVersionUID = -1644016734454696766L;
 
+	public Identities() {
+		super();
+	}
+
 	/**
 	 * Get an unmodifiable list of stored instances of Identity
 	 * @return unmodifiable set of Identity

```

Is needed to persist the Contacts objects.

julianseeger:
done

PR #453, opened 2016-02-10
#451 don't reuse HMac on CryptoUtils to make its usage thread safe, too

julianseeger:
see #451 for details

srsly don't know how to test this as the margin for threading issues is the `hmac.update` itself and the `write` to a local memorybased stream.

+fixed code style issues mentioned by @L-Henke


L-Henke:
This actually seems very hard to test, but I think it will not be necessary, because it is clearly better than before and consistent with the local GCMCipher.

PR #455, opened 2016-02-11
modified AccountingHttp for new prefix resource

julianseeger:
Qabel/qabel-accounting#43

needed since Qabel/qabel-accounting#45


audax:
gson is amazing…

Issue #442, opened 2016-01-18
Private key operations in KeyPair Class


roeslpa:
In order to ensure that the private key is never misused I would suggest to process all operations with the private key in the QblECKeyPair Class and thus remove the public method getPrivateKey(). As far as I know only the computation of the index DM name uses the private key directly. Any arguments against?

roeslpa:
@jmt85 why was this issue closed? It might be implemented in the future, but in my view it is relevant.  If you disagree, please provide a reason.

audax:
The issue of accidently misusing the private key is tackled by code reviews. Moving all operations with the private key into the class couples it very deeply with all other modules.

See julians answer: https://github.com/Qabel/qabel-core/pull/467#issuecomment-194487288

We are currently refactoring contacts, identities and all that stuff, including the persistance. After that refactoring a clean solution for how to couple the private key to modules that use the private key may present itself.

roeslpa:
From my point of view there are only very few use cases for the private key like decryption, signing, being input for a finger print (or hash sum for e.g., the prefix) and being exported. Since all this can be implemented totally independent from modules inside the private key class, there would not be a need for code review of external modules to ensure that it is not misused.

I am not writing this to complicate the implementation but there are several attack vectors which could arise by using the key (e.g.: timing issues) and by preventing the usage outside the class a review would be much easier.

julianseeger:
If it keeps independent of lower level concepts ( unlinke #467 ) I'm totally ok with pulling the actions into one key class as long as is represents a single responsibility (*crypting / hashing /... stuff).

Don't forget that we have to persist and export the private key => some parts just need to get it. And it might give a wrong impression of security because you can always use reflections like
```JAVA
Field key = QblEcKeyPair.class.getField("privateKey");
key.setAccessible(true);
byte[] owned = key.getValue(keyPair);   // or similar...
```

PR #467, opened 2016-03-09
Added index DM name calculation

roeslpa:
Added index DM name calculation to key class to reduce need of private key usage outside this class (see #442 )

### DIFF
--- a/src/main/java/de/qabel/core/crypto/QblECKeyPair.java
+++ b/src/main/java/de/qabel/core/crypto/QblECKeyPair.java
@@ -71,26 +71,9 @@ public class QblECKeyPair implements Serializable {
 		return pubKey;
 	}
 
+	// TODO: Shift all usages of the private key to this class and export the key only password protected
 	public byte[] getPrivateKey() {
 		return privateKey;
-	}
-
-	/**
-	 * Calculates the name of the index DM for a given prefix
-	 * @param prefix Name of the prefix
-	 * @return Name of the index DM
-	 */
-	public String getRootRef(String prefix) {
-		SHA256Digest md = new SHA256Digest();
-		// Byte array for result of hash of 256 Bit = 256/8 Byte
-		byte[] digest = new byte[md.getDigestSize()];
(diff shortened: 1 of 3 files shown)


src/main/java/de/qabel/core/crypto/QblECKeyPair.java
```diff
@@ -71,6 +75,24 @@ public QblECPublicKey getPub() {
...
 
+	/**
+	 * Calculates the name of the index DM for a given prefix
+	 * @param prefix Name of the prefix
+	 * @return Name of the index DM
+	 */
+	public String getRootRef(String prefix) {
+		SHA256Digest md = new SHA256Digest();
+		// Byte array for result of hash of 256 Bit = 256/8 Byte
+		byte[] digest = new byte[256/8];
```
julianseeger: what about `new byte[md.getDigestSize()]`? It would clear the intent and remove the magic number (still magic no matter how many comments there are ^^)

julianseeger:
Can you think of a reasonable test?
Something like "take this [...] prefix and this [...] private key and expect this [...] rootRef"? We could even include such examples in the spec for reference

roeslpa:
Since we should expect the correctness of the digest, your proposal is the only useful test I could think of..

julianseeger:
Then I would like to see one for the sake of having that code tested because it can break a lot.

Btw.: I didn't know about #442 but imho this PR violates a lot of concepts. The QblEcKeyPair is a core concept concerning identities and has nothing to do with Qabel Box itself. Thus, it shouldn't know about Prefixes which are a even a technical detail of the Box module (if you wan't to call it module). It is so "core" that is shouldn't even know about drop.
Box components should depend on the keys and not reverse.

If you wan't to address #442 and are concerned of accidential misuses of the keys itself, I would propose making them protected or package local. Then you could add context-specific subclasses / proxies / adapters in the same package that add module specific methods. Like a `BoxQblECKeyPair` (the name `QblECKeyPair` is bad btw) that provides the `getRootRef` method or a `DropQblECKeyPair` providing Drop specific methods etc.

PR #518, opened 2016-06-16
Convert to java 1.7 + kotlin code

audax:
To make it possible to use the code in Android, this PR converts the core code to 1.7 and Kotlin (which is byte code compatible with 1.6)


audax:
This is stupid because Kotlin doesn't like method names like "getFoo" and "setFoo" that are not properties.

audax:
@julianseeger Please ensure that this version also works in qabel-desktop and maybe do something about the hardcoded block-server version.

audax:
Block server fixed in https://github.com/Qabel/qabel-block/pull/56

audax:
I removed the hardcoded block server commit

julianseeger:
This is truely impossible to review. I'm going to check it against the desktop-client and if it works (combined with the green build), it should be good enough ^^

audax:
Well, I did not change any functionality, at least not intentionally. If this code still works fine with the clients and did not introduce any regressions, it should be fine.

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -0,0 +1,596 @@
...
+                                  protected val readBackend: StorageReadBackend, protected val writeBackend: StorageWriteBackend) : BoxNavigation {
+    private val scheduler = Executors.newScheduledThreadPool(1)
+    protected val cryptoUtils: CryptoUtils
+
+    private val deleteQueue = HashSet<String>()
+    private val changes = LinkedList<DirectoryMetadataChange<out Any>>()
+    open protected var indexNavigation: IndexNavigation? = null
+
+    private var autocommit = true
+    private var autocommitDelay = DEFAULT_AUTOCOMMIT_DELAY
```
julianseeger: Needs to stay public until you implement that injectable value object that controls this in a senseful manner ^^

box/src/main/java/de/qabel/box/storage/DirectoryMetadata.java
```diff
@@ -1,528 +0,0 @@
...
-import java.security.NoSuchAlgorithmException;
-import java.sql.*;
-import java.util.ArrayList;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.UUID;
-import java.util.concurrent.Callable;
-
-public class DirectoryMetadata extends AbstractMetadata {
-    public static final long DEFAULT_SIZE = 56320L;
```
julianseeger: we need that somewhere

box/src/main/java/de/qabel/box/storage/DirectoryMetadata.kt
```diff
@@ -0,0 +1,568 @@
...
+            } catch (e: SQLException) {
+                throw QblStorageException(e)
+            }
+
+        }
+        return AbstractMetadata.TYPE_NONE
+    }
+
+    companion object {
+        val DEFAULT_SIZE = 56320L
```
julianseeger: Ok it wasn't a beauty but now it is
```JAVA
DirectoryMetadata.Companion.getDEFAULT_SIZE()
```
that is wrong in so many ways ^^

box/src/main/java/de/qabel/box/storage/FileMetadata.kt
```diff
@@ -0,0 +1,106 @@
...
+                    val key = rs.getBytes(++i)
+                    return BoxExternalFile(QblECPublicKey(ownerKey), prefix, block, name, size, mtime, key)
+                }
+                return null
+            }
+        } catch (e: SQLException) {
+            throw QblStorageException(e)
+        }
+
+    companion object {
```
julianseeger: We leak the fact that we are using Kotlin... `FileMetadata.Companion.openExisting(...)`
To quote you:
> We leak like the titanic

julianseeger:
The exception handling is f***** up. The kotlin based code throws exceptions that Java does not understand.
```
java.lang.NoClassDefFoundError: kotlin/TypeCastException
	at de.qabel.box.storage.BoxVolume.createIndex(BoxVolume.java:117)
	at de.qabel.box.storage.BoxVolume.createIndex(BoxVolume.java:109)
	at de.qabel.desktop.cellValueFactory.BoxObjectCellValueFactoryTest.setUp(BoxObjectCellValueFactoryTest.java:62)
```
(not sure currently, what caused that exception but it shouldn't result in a `NoClassDefFoundError`)

audax:
I'll remove the Companion-stuff by just throwing "@JVMField" annotations at them.

julianseeger:
Thx. I got 101 of these TypeCastExceptions ...

julianseeger:
oh lol and I found a test for the ProgressInputStream from the box component ... it didn't survive a package move obviously:
```
public class ProgressInputStreamTest {
    @Test
    public void notifiesAboutReads() throws IOException {
        List<Long> updates = new LinkedList<>();
        ByteArrayInputStream in = new ByteArrayInputStream("+".getBytes());
        ProgressInputStream sut = new ProgressInputStream(in, updates::add);
        try {

            byte[] b = new byte[1];
            sut.read(b);
            assertEquals("+", new String(b));
            AsyncUtils.waitUntil(() -> updates.size() == 1);
            assertEquals(1L, (long) updates.get(0));
        } finally {
            sut.close();
        }
    }

    @Test
    public void notifiesOnClose() throws IOException {
        List<Long> updates = new LinkedList<>();
        ByteArrayInputStream in = new ByteArrayInputStream("+".getBytes());
        ProgressInputStream sut = new ProgressInputStream(in, updates::add);
        try {
            sut.close();
            AsyncUtils.waitUntil(() -> updates.size() == 1);
            assertEquals(0L, (long) updates.get(0));
        } finally {
            sut.close();
        }
    }
}
```

audax:
The Exception handling is not fucked up. It throws a NoClassDeffFoundError because the Kotlin stdlib is not included, so the Exception is not found.

The exception itself is probably a null value or something like that.

julianseeger:
Then it should be included in the qabel-core jar ?!
There are also places where it can't be a null value like ProgressInputStream::read
```KOTLIN
public open fun read(): kotlin.Int { /* compiled code */ }
```
```
java.lang.NoClassDefFoundError: kotlin/jvm/internal/Intrinsics
	at de.qabel.box.storage.ProgressInputStream.<init>(ProgressInputStream.kt)
	at de.qabel.desktop.daemon.management.ProgressInputStreamTest.notifiesAboutReads(ProgressInputStreamTest.java:19)
```

audax:
Yes, the stdlib should be included in the qabel-core jar or we switch to using a maven repository.

julianseeger:
You could also add it to the readme until that magic day when we do stuff with a maven repo.
When it's not even in the readme, where should a user (like me) not that from? Magic?

julianseeger:
this is a blocker. what's left before we can merge this?

PR #521, opened 2016-07-04
Repositories

audax:
Moved from android and desktop


core/src/main/java/de/qabel/core/StringUtils.java
```diff
@@ -0,0 +1,12 @@
+package de.qabel.core;
+
+import java.util.List;
+
+public class StringUtils {
+    public static String join(String separator, String[] parts) {
+        return org.apache.commons.lang3.StringUtils.join(parts, separator);
```
julianseeger: static import

julianseeger:
see https://github.com/Qabel/qabel-desktop/pull/393

PR #525, opened 2016-07-12
Identity 256

taynaps:
#256 added notify observers in the identity


core/src/main/java/de/qabel/core/config/IdentityObserver.java
```diff
@@ -0,0 +1,6 @@
+package de.qabel.core.config;
+
+public interface IdentityObserver {
+    public abstract void update();
```
julianseeger: as it's an interface now, `public` and `abstract` can (and should) be omitted

julianseeger:
Looks good to me. As soon as the test are green, this can be merged

PR #524, opened 2016-07-11
File hashes

julianseeger:
resolves #493

PR #526, opened 2016-07-13
Stream API for AbstractNavigation

julianseeger:
depends on https://github.com/Qabel/qabel-core/pull/524


julianseeger:
this is imho done with https://github.com/Qabel/qabel-core/pull/526/commits/b13b42b8144d3334c2b669dccf8250ea261caddc

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -43,13 +43,18 @@ abstract class AbstractNavigation(
     private var autocommit = true
     private var autocommitDelay = DEFAULT_AUTOCOMMIT_DELAY
     private var lastAutocommitStart: Long = 0
+    var time: () -> Long = { System.currentTimeMillis() }
```
audax: Why is is not a method?

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -159,35 +157,43 @@ abstract class AbstractNavigation(
...
+    override fun upload(name: String, file: InputStream, size: Long, listener: ProgressListener?)
+        = uploadStream(file, name, size, time(), listener)
 
     @Synchronized @Throws(QblStorageException::class)
     override fun overwrite(name: String, file: File, listener: ProgressListener?): BoxFile
         =  uploadFile(name, file, null, listener)
 
     @Throws(QblStorageException::class)
     private fun uploadFile(name: String, file: File, expectedFile: BoxFileState?, listener: ProgressListener?): BoxFile {
+        val mtime: Long
```
audax: try/catch is an expression:
`val a: Int? = try { parseInt(input) } catch (e: NumberFormatException) { null }`

box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
```diff
@@ -140,6 +144,28 @@ public void testDeleteFile() throws QblStorageException, IOException {
         nav.download(boxFile);
     }
 
+    @Test
+    public void uploadsStreams() throws Exception {
+        InputStream in = new ByteArrayInputStream("testContent".getBytes());
+        Long size = 11L;
+
+        DefaultIndexNavigation nav = volume.navigate();
+        nav.setTime(new Function0<Long>() { @Override public Long invoke() { return 1234567890L; }}); // imagine lambda
```
audax: this is why it is not a method, I see…

It looks weird, though.

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
julianseeger: ? it's a callback ... Imagine this as a 
```JAVA
@FunctionalInterface
class DefaultTimeProvider implements TimeProvider {
    @Override
    public Long getTime() {
        return System.currentTimeMillis();
    }
}
```

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
julianseeger: hats actually more code ;) but I'll try

julianseeger:
AFTER this PR, I'm going to convert the test class to kotlin ... it looks this ugly because of the missing lambdas / functional interfaces at this level. But konverting it inside this PR would obfuscate too much

box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
julianseeger: image
btw

audax: That looks better :D

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
audax: A rename to "timeProvider" would be enough to make the intent clear, imo.

julianseeger: true. `time` looked good in the test so I thought that was a good idea back then o.O

PR #527, opened 2016-07-19
#184 - AccountingHttp shows now real value of quota from API

nnice:


core/src/main/java/de/qabel/core/accounting/QuotaDto.java
```diff
@@ -0,0 +1,6 @@
+package de.qabel.core.accounting;
+
+public class QuotaDto {
+    String quota;
+    String size;
+}
```
julianseeger: everything package local or everything public or getter ... and a constructor. I would prefer a constructor and two getters to make clean it's immutable

enkore:
Considering that we moved quota handling to the block server maybe the implementation of this should be in a block-related class?

julianseeger:
At least the name is bad, indeed. It has grown that way when Features were moved from the accounting Server to the block server. But I think the Clients shouldn't care which Service implements them for the time being so rename over splitting.

Any suggestions for a better name? Maybe just BoxClient?

julianseeger:
Try to look at that copy&pasted code again. IF you really just copy and paste (to understand the code, you should at least copy it by reading and typing it) you need to review if it really does what you want.

box/src/test/java/de/qabel/box/storage/BoxVolumeBlockTest.java
```diff
@@ -14,7 +14,7 @@
 
 public class BoxVolumeBlockTest extends BoxVolumeTest {
     private BlockReadBackend readBackend;
-    private static AccountingHTTP accountingHTTP;
+    private static BoxHttpClient accountingHTTP;
```
julianseeger: default to the interface. That way you can make sure no method is missing

core/src/main/java/de/qabel/core/accounting/BoxHttpClient.java
```diff
@@ -86,9 +88,41 @@ public void login() throws IOException, QblInvalidCredentials {
...
+            }
+            HttpEntity entity = response.getEntity();
+            if (entity == null) {
+                throw new IOException("No answer from quota");
+            }
+            String responseString = EntityUtils.toString(entity);
+            try {
+                QuotaDto parsedDto = gson.fromJson(responseString, QuotaDto.class);
+                profile.setQuota(parsedDto.quota);
+                quota = parsedDto.quota;
```
julianseeger: What about the `size`?
For displaying `1,7 GB / 2 GB used`, you need both, quota and size (if I understand correctly, what `size` means). ... Those names are not very distinct imho, but that's not your fault.

core/src/main/java/de/qabel/core/accounting/BoxClient.java
```diff
@@ -0,0 +1,35 @@
...
+import de.qabel.core.exceptions.QblCreateAccountFailException;
+import de.qabel.core.exceptions.QblInvalidCredentials;
+import org.apache.http.HttpRequest;
+import org.apache.http.client.utils.URIBuilder;
+
+import java.io.IOException;
+import java.util.ArrayList;
+
+public interface BoxClient {
+    String EMAIL_KEY = "email";
```
julianseeger: unused?

core/src/main/java/de/qabel/core/accounting/BoxClient.java
```diff
@@ -0,0 +1,35 @@
...
+
+    void login() throws IOException, QblInvalidCredentials;
+
+    QuotaDto getQuota() throws IOException, QblInvalidCredentials;
```
julianseeger: using a class named `Dto` is one thing, but using it in a cross-lib-interface is more like reversed hungarian notation. Doesn't look right. Can't this be a QuotaResponse or QuotaState or QuotaReport ?

core/src/main/java/de/qabel/core/accounting/BoxClient.java
```diff
@@ -0,0 +1,35 @@
...
+
+    void authorize(HttpRequest request) throws IOException, QblInvalidCredentials;
+
+    void updatePrefixes() throws IOException, QblInvalidCredentials;
+
+    void createPrefix() throws IOException, QblInvalidCredentials;
+
+    URIBuilder buildUri(String resource);
+
+    URIBuilder buildBlockUri(String resource);
```
julianseeger: are these useful for the interface? They look like internal URIBuilder

core/src/main/java/de/qabel/core/accounting/BoxClient.java
nnice: sure thats a good point, changing into QuotaState

core/src/main/java/de/qabel/core/accounting/BoxHttpClient.java
```diff
@@ -86,11 +88,43 @@ public void login() throws IOException, QblInvalidCredentials {
...
+        try (CloseableHttpResponse response = httpclient.execute(httpGet)) {
+            int statusCode = response.getStatusLine().getStatusCode();
+            if (statusCode >= 400) {
+                throw new IllegalStateException(
+                    "Server responded with " + statusCode + ": " + response.getStatusLine().getReasonPhrase()
+                );
+            }
+            HttpEntity entity = response.getEntity();
+            if (entity == null) {
+                throw new IOException("No answer from quota");
```
julianseeger: from blockserver on quota request? or what is this quota? ;)

nnice: the quota request did not answered there. well okay quota request lives on the blockeserver

core/src/main/java/de/qabel/core/accounting/BoxHttpClient.java
```diff
@@ -25,27 +25,29 @@
 import java.net.URISyntaxException;
 import java.util.*;
 
-public class AccountingHTTP {
-
-    private static final Logger logger = LoggerFactory.getLogger(AccountingHTTP.class.getName());
-    public static final String EMAIL_KEY = "email";
+public class BoxHttpClient implements BoxClient {
+    String EMAIL_KEY = "email";
```
julianseeger: private

core/src/main/java/de/qabel/core/accounting/BoxHttpClient.java
```diff
@@ -25,27 +25,29 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(BoxHttpClient.class.getName());
 
     private AccountingServer server;
     private final CloseableHttpClient httpclient;
     private Gson gson;
     private AccountingProfile profile;
 
-    public AccountingHTTP(AccountingServer server, AccountingProfile profile) {
+    public QuotaState quotaState;
```
julianseeger: private

core/src/main/java/de/qabel/core/accounting/BoxHttpClient.java
```diff
@@ -86,11 +88,42 @@ public void login() throws IOException, QblInvalidCredentials {
...
+                    "Server responded with " + statusCode + ": " + response.getStatusLine().getReasonPhrase()
+                );
+            }
+            HttpEntity entity = response.getEntity();
+            if (entity == null) {
+                throw new IOException("No answer from quotaState");
+            }
+            String responseString = EntityUtils.toString(entity);
+            try {
+                quotaState = gson.fromJson(responseString, QuotaState.class);
```
julianseeger: why is this a class property?

core/src/main/java/de/qabel/core/accounting/QuotaState.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public long getQuota() {
+        return quota;
+    }
+
+    public long getSize() {
+        return size;
+    }
+
+    public String getQuotaDescription() {
```
julianseeger: This is view logic. Rename it to `toString` to make clear it's a debug only feature or move it completely to a UI related class (in the client)

core/src/main/java/de/qabel/core/accounting/QuotaState.java
```diff
@@ -0,0 +1,20 @@
+package de.qabel.core.accounting;
+
+import org.apache.commons.io.FileUtils;
+
+class QuotaState {
+    public long quota;
+    public long size;
```
julianseeger: private properties + constructor

PR #530, opened 2016-07-22
Chat drop message

JoGir:


core/build.gradle
```diff
@@ -45,8 +45,8 @@ dependencies {
 	compile group: 'org.apache.logging.log4j', name: 'log4j-slf4j-impl', version: '2.4'
 	compile group: 'org.apache.james', name: 'apache-mime4j', version: '0.7.2'
 	compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.+'
+    compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.2'
```
enkore: nag: indentation

core/src/main/java/de/qabel/core/repository/GenericEntityManager.java
```diff
@@ -36,6 +36,11 @@ public boolean contains(Class entityType, I id) {
         return (T)entities.get(entityType).get(id);
     }
 
+
+    public <T> void remove(Class<T> entityType, Integer id){
+        entities.get(entityType).remove(id);
```
julianseeger: what is this for? shouldn't be neccessary. New Entities should get a new ID and old entities are garbage collected anyways

core/src/main/java/de/qabel/core/repository/entities/ChatDropMessage.kt
```diff
@@ -0,0 +1,27 @@
...
+        NEW(0), READ(1), PENDING(2), SENT(3);
+    }
+
+    enum class MessageType(val type: String) {
+        SHARE_NOTIFICATION("box_share_notification"),
+        BOX_MESSAGE("box_message")
+    }
+
+    enum class Direction(val type: Int) {
+        INCOMING(0), OUTGOING(1)
```
julianseeger: what is this int? do you mean true/false? ... then it should be possible to use true/false

core/src/main/java/de/qabel/core/repository/entities/ChatDropMessage.kt
```diff
@@ -0,0 +1,27 @@
...
+                           val contactId: Int,
+                           var identityId: Int,
+                           val direction: Direction,
+                           val status: Status,
+                           val type: MessageType,
+                           var payload: String,
+                           val createdOn: Long) : BaseEntity {
+
+    enum class Status(val type: Int) {
+        NEW(0), READ(1), PENDING(2), SENT(3);
```
julianseeger: the constructor arguments (0/1/2/3) are optional. what are they for? internal database identifier?

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
```diff
@@ -0,0 +1,322 @@
...
+        val contactsIDs = contacts.map { contact -> contact.id }
+
+        val contactsDropUrls = dropUrlRepository.findDropUrls(contactsIDs)
+        val contactIdentityKeys = findContactIdentityKeys(contactsIDs)
+
+        return contacts.map { contact ->
+            contactsDropUrls[contact.id]?.forEach { dropUrl -> contact.addDrop(dropUrl) }
+            Pair(contact,
+                contactIdentityKeys.mapEntities(contact.id, identities))
+        }
```
julianseeger: please don't build up your contacts in this method. Thats what the Hydrators do (at least it's the current structure in all repository and nobody complained about that yet).
Take a look at the IdentityHydrator. It does something similar for the identities (fetching dropUrls etc.).  Maybe cou can reuse that code. Imho the code in this method is in no way obvious and thus very hard to read. Even a plain SQL statement would be clearer even though reusing the existing methods better in general.

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+import de.qabel.core.repository.entities.DropState
+import de.qabel.core.repository.exception.EntityNotFoundException
+import de.qabel.core.repository.framework.BaseRepositoryImpl
+import de.qabel.core.repository.sqlite.schemas.DropStateDB
+
+class SqliteDropStateRepository(database: ClientDatabase,
+                                entityManager: EntityManager) :
+    BaseRepositoryImpl<DropState>(DropStateDB, database, entityManager), DropStateRepository {
+
+    override fun getDropState(dropId: String) = findByDropId(dropId, true)!!.eTag
```
julianseeger: Look at the interface. It is supposted to throw an EntityNotFoundException if the entity is not found. This implementation throws a NullPointerException. Not what kotlin was intended for ;)

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+    private fun findByDropId(dropId: String, forceNotNull: Boolean): DropState? {
+        try {
+            val query = createEntityQuery()
+            query.whereAndEquals(DropStateDB.DROP, dropId);
+            return getSingleResult(query)
+        } catch(ex: EntityNotFoundException) {
+            if (forceNotNull) {
+                throw ex;
+            }
+            return null;
```
julianseeger: EntityNotFoundException

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    override fun setDropState(dropId: String, state: String) =
+        DropState(dropId, state).let { persist(it) }
```
julianseeger: setDropState is void => 
```kotlin
override fun setDropState(dropId: String, state: String) = persist(DropState(dropId, state))
```
don't add unneccessary komplexity

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropUrlRepository.java
```diff
@@ -66,4 +70,30 @@ public void store(Entity contact, int contactId) throws SQLException {
             }
         }
     }
+
+    public Map<Integer, List<DropURL>> findDropUrls(List<Integer> contactIds) throws PersistenceException {
```
julianseeger: just hydrate the contacts with their dropUrls... then you don't need this method

core/src/main/java/de/qabel/core/repository/sqlite/hydrator/SimpleContactHydrator.kt
```diff
@@ -0,0 +1,45 @@
...
+
+        recognize(contact)
+        return contact
+    }
+
+    override fun recognize(instance: Contact) {
+        entityManager.put(Contact::class.java, instance)
+    }
+
+    override fun getFields(): Array<out String> {
```
julianseeger: feel free to extend the AbstractHydrator (or to convert it to a delegation). If you introduce a concept like the relation, try to apply it to the whole codebase where neccessary. It will give you a good insight where it doesn't fit.

core/src/main/java/de/qabel/core/repository/sqlite/migration/Migration1460997040ChatDropMessage.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public long getVersion() {
+        return 1460997040L;
+    }
+
+    @Override
+    public void up() throws SQLException {
+        execute(
+            "CREATE TABLE chat_drop_message (" +
```
julianseeger: is this chat specific? All drop messages are stored here, aren't they?

core/src/main/java/de/qabel/core/repository/sqlite/schemas/ContactDB.kt
```diff
@@ -0,0 +1,17 @@
+package de.qabel.core.repository.sqlite.schemas
+
+
+object ContactDB {
```
julianseeger: why is this not a DBRelation?

core/src/main/java/de/qabel/core/util/DefaultHashMap.kt
```diff
@@ -0,0 +1,18 @@
...
+class DefaultHashMap<K,V>(private val valueFactory: (K) -> V) : HashMap<K, V>() {
+
+    @Synchronized override operator fun get(key: K): V? {
+        if (!containsKey(key)) {
+            put(key, valueFactory(key))
+        }
+        return super.get(key)
+    }
+
+    fun getOrDefault(key: K): V {
```
julianseeger: this method applies no default. It's a `getOrNullPointer`

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
JoGir: Nope, there is no NullPointer...

core/src/main/java/de/qabel/core/util/DefaultHashMap.kt
JoGir: Nope, there is no NullPointer...

core/src/main/java/de/qabel/core/repository/sqlite/schemas/ContactDB.kt
JoGir: Older code from Android, but ContactRepo isnt a Baserepository...

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
julianseeger: What is that forceNotNull for ? it's private and as far as I can tell only used in getDropState with `true`

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropUrlRepository.java
```diff
@@ -66,4 +70,30 @@ public void store(Entity contact, int contactId) throws SQLException {
...
+            Map<Integer, List<DropURL>> contactDropUrlMap = new HashMap<>();
+            try (PreparedStatement statement = database.prepare(
+                "SELECT contact_id, url " +
+                    "FROM " + TABLE_NAME + " urls " +
+                    "WHERE contact_id IN (" + StringUtils.join(",", contactIds) + ")"
+            )) {
+                try (ResultSet results = statement.executeQuery()) {
+                    while (results.next()) {
+                        int id = results.getInt(1);
+                        List<DropURL> identityKeys = contactDropUrlMap.get(id);
```
julianseeger: is this copy&paste code or why are the drop urls stored in list called `identityKeys`? Don't understand the meaning.

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropUrlRepository.java
```diff
@@ -66,4 +70,30 @@ public void store(Entity contact, int contactId) throws SQLException {
...
+                        if (identityKeys == null) {
+                            identityKeys = new LinkedList<>();
```
julianseeger: your default hash map should work here shouldn't it?

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropUrlRepository.java
```diff
@@ -82,12 +84,13 @@ public void store(Entity contact, int contactId) throws SQLException {
                 try (ResultSet results = statement.executeQuery()) {
                     while (results.next()) {
                         int id = results.getInt(1);
-                        List<DropURL> identityKeys = contactDropUrlMap.get(id);
-                        if (identityKeys == null) {
-                            identityKeys = new LinkedList<>();
-                            contactDropUrlMap.put(id, identityKeys);
-                        }
-                        identityKeys.add(new DropURL(results.getString(2)));
+                        List<DropURL> dropURLList = contactDropUrlMap.getOrDefault(id, new CheckedFunction<Integer, List<DropURL>>() {
```
julianseeger: o.O konvert to kotlin too? XD

core/src/main/java/de/qabel/core/util/DefaultHashMap.kt
audax: This should be switched to a Delegate instead of inheritance with only the getOrDefault method added. I think that's the only way it is used anyway.

core/src/main/java/de/qabel/core/http/MainDropConnector.kt
```diff
@@ -0,0 +1,54 @@
...
+        if (!dropResult.second.isEmpty()) {
+            dropState.eTag = dropResult.second
+        }
+        val dropMessages = mutableListOf<DropMessage>()
+        for (byteMessage in dropResult.third) {
+            val binaryFormatVersion = byteMessage[0]
+            val binaryMessage: AbstractBinaryDropMessage? = when (binaryFormatVersion) {
+                0.toByte() -> {
+                    var m: BinaryDropMessageV0? = null
+                    try {
```
audax: use try as an expression instead.

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
```diff
@@ -31,7 +31,7 @@ class SqliteContactRepository(database: ClientDatabase,
         SqliteDropUrlRepository(database, DropURLHydrator())), ContactDB.TABLE_NAME), ContactRepository {
 
     @Throws(PersistenceException::class, EntityNotFoundException::class)
-    internal fun find(id: Int?): Contact {
+    override fun find(id: Int?): Contact {
```
audax: How would a find with id=null work?

core/src/main/java/de/qabel/core/repository/sqlite/SqliteDropStateRepository.kt
audax: forcoNotNull throws an exception instead of returning null. The forceNotNull flag should be removed.

An extension method can do that job just fine.

core/src/main/java/de/qabel/core/repository/framework/BaseRepositoryImpl.kt
```diff
@@ -53,14 +53,14 @@ abstract class BaseRepositoryImpl<T : BaseEntity>(val relation: DBRelation<T>,
         client.prepare(deleteStatement).use {
             it.setInt(1, id)
             it.execute()
-            entityManager.remove(relation.ENTITY_CLASS, id)
         }
     }
 
     fun findById(id: Int): T {
-        if (entityManager.contains(relation.ENTITY_CLASS, id)) {
+        //TODO Cant use entityManager as Cache
```
audax: uhm...why?

core/src/main/java/de/qabel/core/repository/sqlite/SqliteChatDropMessageRepository.kt
```diff
@@ -23,39 +23,39 @@ class SqliteChatDropMessageRepository(val database: ClientDatabase,
 
     override fun findByContact(contactId: Int, identityId: Int): List<ChatDropMessage> {
         val queryBuilder = createEntityQuery()
-        queryBuilder.whereAndEquals(CONTACT_ID, contactId)
-        queryBuilder.whereAndEquals(IDENTITY_ID, identityId)
-        queryBuilder.orderBy(ChatDropMessageDB.CREATED_ON.exp())
+        .whereAndEquals(CONTACT_ID, contactId)
```
audax: Why not just a with(createEntityQuery()) instead of all that "return this"?

PR #531, opened 2016-07-22
Docker

### LINKED ISSUE #529: Update Vagrantfile to use Docker images. Purge everything else.
Also see https://github.com/Qabel/qabel-infrastructure/issues/1

enkore:
Fixes #529 

Tested with the infra tests, works there. If debugging is needed commands can be run through docker as usual, but I'll also try experimenting with SSHish stuff if there's interest.

NOTE: There is one teeny weeny giant MSA dependency here, the port configuration. So this still has to be updated if a new (public) service is added.


enkore:
So after some fiddling with java-gradle stuff building with "./gradlew build" seems to actually work here now. It builds and passes all tests with this new container/vagrant. Only `vagrant up` is needed (as before).

julianseeger:
Doesn't work for me. The container dies after a couple of seconds telling 
`django.db.utils.OperationalError: FATAL:  the database system is starting up`
maybe a timing issue? [full trace here](https://gist.github.com/julianseeger/7b21549a0c9faff6dc7080207af5f527)

julianseeger:
The new container works for me with all core tests and desktop tests.
To reduce the duplication of portdefinitions, you could extract them to one list, it's just ruby at last. (and for non-docker-in-virtualbox machines the "forwarded_ports" stuff is optional)

enkore:
ftr the fixing commit is qabel/qabel-infrastructure@4efd234

julianseeger:
I think the forwarded_ports is required if you you can't run docker natively and a VM is spawned to run docker inside it. Not sure about that.

PR #534, opened 2016-07-25
make BoxVolumeImpl.loadDriver() great again

audax:
internal methods are not visible in other modules like qabel-android.


audax:
I'll deliver a more complete PR with all the stuff that android needs.

PR #536, opened 2016-07-26
Index server API

### LINKED ISSUE #498: communication between core and index server
For the upload function in android and the desktop client should a common function in the core implemented which handles the communication to the server reg. and the clients.

enkore:
Very WIP. Search is complete, updating not at all there yet.

I invite everyone to take a peek and point out problems, suggestions etc :)

Fixes #498


core/src/main/java/de/qabel/core/index/IndexHTTP.kt
```diff
@@ -0,0 +1,27 @@
+package de.qabel.core.index
+
+import com.google.gson.Gson
+import org.apache.http.impl.client.CloseableHttpClient
+import org.apache.http.impl.client.HttpClients
+
+class IndexHTTP(
```
enkore: interface required

core/src/main/java/de/qabel/core/index/IndexHTTP.kt
```diff
@@ -0,0 +1,27 @@
...
+    val server: IndexServer,
+    val httpClient: CloseableHttpClient = HttpClients.createMinimal()
+) {
+    private val gson: Gson = createGson()
```
enkore: no

core/src/main/java/de/qabel/core/index/IndexHTTP.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    fun search(attributes: Map<FieldType, String>): List<PublicIdentity> {
+        val search = Search(server, gson)
```
enkore: inject through constructor w/ default value

core/src/main/java/de/qabel/core/index/IndexServer.kt
```diff
@@ -0,0 +1,17 @@
+package de.qabel.core.index
+
+import org.apache.http.client.utils.URIBuilder
+import java.net.URI
+
+data class IndexServer(val indexUri: URI) {
```
enkore: rename

core/src/main/java/de/qabel/core/index/Search.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import java.net.URISyntaxException
+import java.util.*
+
+
+internal class Search(
+    private val server: IndexServer,
+    private val gson: Gson,
+    private val logger: Logger = LoggerFactory.getLogger(Search::class.java)
```
enkore: lazy attribute in body

core/src/main/java/de/qabel/core/index/models.kt
```diff
@@ -0,0 +1,125 @@
...
+    val keyPair: QblECKeyPair,
+    val dropURL: DropURL,
+    val alias: String
+) {
+    companion object {}
+}
+
+fun IdentityData.Companion.fromIdentity(identity: Identity): IdentityData {
+    return IdentityData(identity.primaryKeyPair, identity.helloDropUrl, identity.alias)
+}
```
enkore: move to body of companion object

core/src/main/java/de/qabel/core/index/models.kt
```diff
@@ -0,0 +1,125 @@
...
+enum class FieldType(val fieldName: String) {
+    EMAIL("email"),
+    PHONE("phone"),
+}
+
+
+internal enum class UpdateItemAction(val actionName: String) {
+    CREATE("create"),
+    DELETE("delete"),
+}
```
enkore: review w/ custom field (actionName/fieldName) is necessary or whether internal name field can be overriden.

core/src/main/java/de/qabel/core/index/models.kt
```diff
@@ -0,0 +1,125 @@
...
+                    ))
+                }
+                QblECPublicKey(Hex.decode(it.json.string))
+            }
+            serialize {
+                it.src.readableKeyIdentifier.toJson()
+            }
+        }
+        .create()
+}
```
enkore: research if this can be done some other way, say, annotations

core/src/test/java/de/qabel/core/index/SearchTest.kt
```diff
@@ -0,0 +1,104 @@
...
+            "{", /* that will stay with you all day */
+            "{}",
+            "{\"1identity\": []}"
+        )
+        fun assertBrokenResponse(json: String) {
+            try {
+                search.parseSearchResponse(json)
+            } catch (e: MalformedResponseException) {
+                return
+            }
```
enkore: hamcrest matcher thing

or, fail() (not asserTrue(…, false)

enkore:
Blocked by Qabel/qabel-index#22 for now, switching tasks...

core/src/main/java/de/qabel/core/config/Identity.java
```diff
@@ -142,6 +142,14 @@ public QblECPublicKey getEcPublicKey() {
...
     }
 
+    /**
+     * Returns the drop URL that should be used for the HELLO protocol, i.e.
+     * made public and exported to qabel-index.
+     */
+    public DropURL getHelloDropUrl() {
+        return (new ArrayList<>(getDropUrls())).get(0);
+    }
+
```
enkore: maybe-design-xxx

core/build.gradle
```diff
@@ -39,6 +39,7 @@ dependencies {
 	testCompile group: 'org.meanbean', name: 'meanbean', version: '2.+'
     testCompile 'org.hamcrest:java-hamcrest:2.0.0.0'
 	compile 'com.google.code.gson:gson:2.+'
+	compile 'com.github.salomonbrys.kotson:kotson:2.3.0'
```
enkore: dep screened, is ok (MIT license).

core/src/main/java/de/qabel/core/crypto/QblECPublicKeyIndexTypeAdapter.kt
```diff
@@ -0,0 +1,25 @@
...
+class QblECPublicKeyIndexTypeAdapter : TypeAdapter<QblECPublicKey>() {
+    override fun write(out: JsonWriter, value: QblECPublicKey) {
+        out.value(Hex.toHexString(value.key))
+    }
+
+    override fun read(input: JsonReader): QblECPublicKey {
+        val hexString = input.nextString()
+        val requiredLength = QblECPublicKey.KEY_SIZE_BYTE * 2
+        if (hexString.length != requiredLength) {
+            /* XXX Shouldn't QblECPublicKey have this check? */
```
enkore: can't really insert a new exception in QblECPublicKey since it's used everywhere - would have to check all call-sites 'till the top of the tree.

enkore:
TODO: complete tests, write docs.

core/src/main/java/de/qabel/core/index/IndexHTTPLocation.kt
```diff
@@ -0,0 +1,17 @@
+package de.qabel.core.index
+
+import org.apache.http.client.utils.URIBuilder
+import java.net.URI
+
+data class IndexHTTPLocation(val indexUri: URI) {
+
+    constructor(indexUri: String) : this(URI(indexUri))
+
+    fun getUriBuilderForEndpoint(endpoint: String): URIBuilder {
```
enkore: internal

core/src/main/java/de/qabel/core/index/Update.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    fun buildJsonUpdateRequest(identity: IdentityData, action: UpdateItemAction): String {
+        val updateRequest = UpdateRequest(
+            identity = identity.toPublicIdentity(),
+            items = getUpdateItemsForIdentity(identity, action)
+        )
+        return gson.toJson(updateRequest)
+    }
+
+    fun encryptJson(json: String, senderKeyPair: QblECKeyPair, serverPublicKey: QblECPublicKey): ByteArray {
```
enkore: So ATM we don't have a way to send unencrypted requests, which is not a problem for now, since both apps need the identity long before any of this can be used.

enkore:
Q: Is it currently possible to even change things like the email address or phone number attached to an identity? I haven't added an API yet that would easily allow this in one operation (although it can be also argued that it may make sense to keep the old email address listed?).

julianseeger:
> Q: Is it currently possible to even change things like the email address or phone number attached to an identity?

If you mean if there is a GUI in the clients, then at least for the desktop client, there is none yet afaik. The feature to attach one should get added in the next few weeks. The ticket to at least set one initially is included here https://github.com/Qabel/qabel-desktop/issues/364 and the ability to edit one is not even scheduled: https://github.com/Qabel/qabel-desktop/issues/390

core/src/main/java/de/qabel/core/crypto/QblECPublicKeyIndexTypeAdapter.kt
julianseeger: What kind of exception would you like to throw?
You can always throw an IllegalArgumentException unchecked ... would fit semantics imo and you don't have to adjust all call-sites (at least not to compile it)

enkore:
Done / no-wip. Server-tests currently fail since the qabel-index is not yet available on the gocd agents

core/src/main/java/de/qabel/core/crypto/QblECPublicKeyIndexTypeAdapter.kt
enkore: I guess that'd work. Separate PR through (if it shall be done)

core/src/main/java/de/qabel/core/index/APIBase.kt
```diff
@@ -0,0 +1,31 @@
...
+            val statusCode = response.statusLine.statusCode
+            APIError.checkResponse(response)
+            val body = if (response.entity == null && statusCode != 204) {
+                // require a body except for 204 (No Content)
+                throw MalformedResponseException("No response from index server.")
+            } else if (response.entity == null && statusCode == 204) {
+                ""
+            } else { /* statusCode != 204 && response.entity != null */
+                EntityUtils.toString(response.entity, Charsets.UTF_8)
+            }
```
julianseeger: either extract this whole if to a new method like getBody or find a way to get rid of these comments .. like `if (hasNoContentButStatusCodeRequiresOne(...)) { throw ... }`  / preferably shorter

core/src/main/java/de/qabel/core/index/APIBase.kt
```diff
@@ -0,0 +1,31 @@
...
+            return parseResponse(body, response)
+        }
+    }
+
+    fun parseResponse(jsonString: String, response: HttpResponse): T {
+        return parseResponse(jsonString)
+    }
+
+    fun parseResponse(jsonString: String): T {
+        throw NotImplementedError("Implement either parseResponse()")
```
julianseeger: this is pest against cholera, isn't it? you trade a compile-time-error + a refused bequest vs a runtime error

core/src/main/java/de/qabel/core/index/Key.kt
```diff
@@ -0,0 +1,8 @@
+package de.qabel.core.index
+
+import de.qabel.core.crypto.QblECPublicKey
+import org.apache.http.client.methods.HttpUriRequest
+
+internal interface Key: APIBase<QblECPublicKey> {
```
julianseeger: I would prefer more descriptive names like `KeyEndpoint` / `KeyAPI` or similar (you can probably find something that fits better if I misunderstood the structure) but this just doesn't seem to be a key. It's some Client-Utility that builds requests and parses responses but not a Key for anything. This makes it imho harder to understand the structure without reading the full diff. key.buildRequest and search.parseResponse. Don't even know what the key endpoint is, probably a CRUD interface for keys on the index server but thats totally guessing

core/src/main/java/de/qabel/core/index/models.kt
```diff
@@ -0,0 +1,133 @@
...
+
+private val PublicIdentityDeserializer = jsonDeserializer {
+    /* By default gson will set fields not contained in the input data to null silently and I see no way
+     * to change that behaviour without writing the required verification by hand. So here it goes.
+     */
+    val obj = it.json.obj
+    if (!obj.contains("public_key") || !obj.contains("alias") || !obj.contains("drop_url")) {
+        throw IllegalArgumentException("missing key in identity")
+    }
+    /* There's no "base serdes" here. If a custom TypeAdapter is around, at least this level has to be spelled out */
```
julianseeger: base serdes?

core/src/main/java/de/qabel/core/index/models.kt
```diff
@@ -0,0 +1,133 @@
...
+import de.qabel.core.crypto.QblECKeyPair
+import de.qabel.core.crypto.QblECPublicKey
+import de.qabel.core.drop.DropURL
+import java.util.*
+
+
+/**
+ * Data class representing a public identity in the index.
+ */
+data class PublicIdentity(
```
julianseeger: Do we really need to introduce another level of metaphorical complexity on top of indentity and contact?
Given that a Contact represents the public contact information and a contact + private information like the private key is the identity ... then semantically a PublicIdentity is a Contact isn't it? Aside from this being a lightweight dto compared to the original Contact class ... I'm totally not complaining about that fact, just the PublicIdentity vs. Contact naming)

core/src/test/java/de/qabel/core/index/SearchTest.kt
```diff
@@ -0,0 +1,133 @@
...
+            """
+            {"identities": [{
+                "public_key": "0f82b0018d1140a37b9cf3a4570bbdd415c8bbbcfc7efe7ef8aa912ce3760520",
+                "drop_url": "http://example.net/somewhere/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopo"
+            }]}
+            """
+        )
+        fun assertIgnoresBrokenIdentity(json: String) {
+            val result = search.parseResponse(json)
+            assertEquals(0, result.size)
```
julianseeger: what happens if one identity is broken but the other ones are not? is the result empty or is only the broken one skipped?

core/src/main/java/de/qabel/core/index/IndexServer.kt
```diff
@@ -0,0 +1,48 @@
+package de.qabel.core.index
+
+
+/**
+ * Index server API
+ *
+ * All methods can raise [IndexServerException] or one of it's subclasses, [APIError] and [MalformedResponseException].
```
julianseeger: I would prefer `@Throws` annotations at least in this interface as it needs to be usable in Java and should imo conform to it by declaring it's exceptions. As far as I understood, `@Throws` annotations are required to to get those exceptions declared on the resulting JVM methods and it doesn't hurt in kotlin.

julianseeger:
Looks good disregarding some minor naming issues and I would like to see `@Throws` annotations on the main interface.
In case we don't adopt `PublicIdentity` as an "official" naming (in the docs, gui etc.) it adds additional confusion to the existing identity-contact-confusion.

core/src/main/java/de/qabel/core/index/APIBase.kt
enkore: KISS - one parseResponse(json, reponse), and people who don't care about the response just don't use it?

enkore:
I think I got most things mentioned covered.

`@Throws` *should* be correct this way, mentioning IOException which is the base class of everything that Apache HC throws (e.g. ClientProtocolException inherits IOException). IndexServerException inherits also from IOException.

>  In case we don't adopt PublicIdentity as an "official" naming (in the docs, gui etc.) it adds additional confusion to the existing identity-contact-confusion.

While the index API spec uses that naming convention, I changed the code here to call it `IndexContact` now, which should be less confusing in this project/context.

julianseeger:
Didn't know that we already use that in the spec, but `IndexContact` sounds logical.

PR #541, opened 2016-08-02
resolves #494 pessimistic folder commit for conflict prevention

### LINKED ISSUE #494: Folder Conflict Handling
to ensure that the error can not occur:

First commit empty folder → empty dm first
If folder already exist, take existing folder
no solution for renaming (ignore second)
=> folder conflict is an idempotent command

julianseeger:
according to #494 , the empty DM gets inserted to the parent-dm which is instantly uploaded, too, to make sure that if it is required to merge two newly created folders, they don't need to be merged because they latter one is empty.

### DIFF
--- a/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
+++ b/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
@@ -398,7 +398,7 @@ abstract class AbstractNavigation(
 
     @Synchronized @Throws(QblStorageException::class)
     override fun createFolder(name: String): BoxFolder {
-        execute(CreateFolderChange(name, folderNavigationFactory, directoryFactory)).boxObject
+        execute(CreateFolderChange(name, folderNavigationFactory, directoryFactory))
         commit()
         refresh()
         return getFolder(name)
@@ -409,7 +409,7 @@ abstract class AbstractNavigation(
 
     @Synchronized @Throws(QblStorageException::class)
     override fun unshare(boxObject: BoxObject) {
-        indexNavigation.getSharesOf(boxObject)?.forEach { share ->
+        indexNavigation.getSharesOf(boxObject).forEach { share ->
             try {
                 indexNavigation.deleteShare(share)
             } catch (e: QblStorageException) {
--- a/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
+++ b/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
@@ -79,7 +79,6 @@ abstract class AbstractNavigation(
         }
     }
 
-
     @Synchronized @Throws(QblStorageException::class)
     override fun commitIfChanged() {
         if (isUnmodified) {
@@ -103,7 +102,7 @@ abstract class AbstractNavigation(
         }
 
         // the remote version has changed from the _old_ version
-        if (updatedDM != null && !Arrays.equals(version, updatedDM.version)) {
+        if (dm !== updatedDM && updatedDM != null && !Arrays.equals(version, updatedDM.version)) {
             logger.info("Conflicting version")
             // ignore our local directory metadata
             // all changes that are not inserted in the new dm are _lost_!
@@ -115,6 +114,11 @@ abstract class AbstractNavigation(
         changes.postprocess(dm, writeBackend)
 
         changes.clear()
+    }
+
+    @Synchronized @Throws(QblStorageException::class)
+    override fun refresh() {
+        dm = reloadMetadata().apply { changes.execute(this) }
     }
 
     override val isUnmodified: Boolean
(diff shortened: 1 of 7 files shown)


box/src/main/java/de/qabel/box/storage/jdbc/JdbcDirectoryMetadata.kt
```diff
@@ -128,6 +129,7 @@ class JdbcDirectoryMetadata(
         md.update(byteArrayOf(0, 1))
         md.update(oldVersion)
         md.update(deviceId)
+        md.update(UUID.randomUUID().toString().toByteArray())
```
julianseeger: Just needed a way to make sure two DMs don't get the version even if they are created on the same device and have the same predecessor. UUID does it's job

box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
```diff
@@ -231,15 +232,15 @@ public void testAutocommitDelay() throws Exception {
...
 
         BoxNavigation nav2 = volume2.navigate();
-        assertFalse(nav2.hasFolder("a"));
+        assertFalse(nav2.hasFile("testfile"));
         waitUntil(new Callable<Boolean>() {
             @Override
             public Boolean call() throws Exception {
                 BoxNavigation nav3 = volume2.navigate();
-                return nav3.hasFolder("a");
+                return nav3.hasFile("testfile");
```
julianseeger: switched from folders to files here because folders always get committed instantaniously

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -392,10 +396,13 @@ abstract class AbstractNavigation(
         }
     }
 
-
     @Synchronized @Throws(QblStorageException::class)
-    override fun createFolder(name: String): BoxFolder
-        = execute(CreateFolderChange(name, folderNavigationFactory, directoryFactory)).boxObject
+    override fun createFolder(name: String): BoxFolder {
+        execute(CreateFolderChange(name, folderNavigationFactory, directoryFactory)).boxObject
```
audax: Either save the boxObject to return it at the end of the function instead of getFolder, or don't access it.

box/src/main/java/de/qabel/box/storage/jdbc/JdbcDirectoryMetadata.kt
audax: The motivation behind the device id being included in the version was exactly that: If the same device made the change, the version should be identical.

Couldn't we now just drop the device id completly or is it used anywhere?

julianseeger: It's still stored in the DM to track "who" authored the last file change / upload.
IIRC the deviceID tracking is some kind of "we could use it in the future" alias yagni ^^

And for the version: don't know why two changes on one device should result in the same version. You would at least need to check if the changes were identical. What happened in my test is that two tests that change the same DM in different ways result in the same new version ... one could argue that the test could change the deviceID but at least on the desktop client where asynchronous stuff happens, two changes to the same DM (not the same object) could occur simultaneously on the same device.

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
julianseeger: yep, leftover. I don't think I can save it for later return because on "conflict", the BoxFolder of the `getFolder`-call would reference the new remote DM instead of the dropped "new" local DM that was known before commit&refresh

box/src/main/java/de/qabel/box/storage/jdbc/JdbcDirectoryMetadata.kt
audax: Indeed, the "author"-stuff was the remaining reason. We had a requirement to show on which client "stuff happened".
I think we should just drop the deviceId from the version hash, but keep it in the DM for now.

box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
```diff
@@ -597,6 +606,43 @@ public void deleteCleansShares() throws Exception {
...
+        subnav2.commit();
+
+        // test the conflict result
+        BoxNavigation nav = volume.navigate();
+        nav.setMetadata(nav.reloadMetadata());
+        assertThat(nav.listFolders(), hasSize(1));
+        BoxNavigation subnav = nav.navigate(folder1);
+        assertTrue(subnav.hasFile("file1"));
+        assertTrue(subnav.hasFile("file2"));
+        assertThat(subnav.listFiles(), hasSize(2));
```
julianseeger: ugly test .. I'm open for suggestions

audax:
Tested locally, runs fine.

PR #545, opened 2016-08-08
Refactoring of DropMessage parsing

audax:
* Identity.toContact() implemented as a utility method
* DropParser from DropConnector extracted
* DropParser returns the identity used to decryptt the message
* No NullPointerExceptions anymore in DropMessage json parsing


core/src/main/java/de/qabel/core/drop/DropSerializer.kt
```diff
@@ -33,15 +33,15 @@ class DropSerializer : JsonSerializer<DropMessage>, JsonDeserializer<DropMessage
...
                 throw JsonParseException("Unexpected version: " + version)
             }
-            val time = it.get(TIME_STAMP).asLong
-            val sender = it.get(SENDER).asString
-            val dropPayload = it.get(PAYLOAD).asString
-            val dropPayloadType = it.get(PAYLOAD_TYPE).asString
+            val time = it.get(TIME_STAMP)?.asLong ?: throw JsonSyntaxException("time is not a long")
+            val sender = it.get(SENDER)?.asString ?: throw JsonSyntaxException("sender not defined")
+            val dropPayload = it.get(PAYLOAD)?.asString ?: throw JsonSyntaxException("payload not defined")
+            val dropPayloadType = it.get(PAYLOAD_TYPE)?.asString ?: throw JsonSyntaxException("payload type not defined")
```
julianseeger: imagine
```KOTLIN
val version = it.safeGet(VERSION).asInt
```
that throws the JsonSyntaxException by it's own ;)

core/src/main/java/de/qabel/core/http/MainDropConnector.kt
```diff
@@ -13,7 +15,9 @@ import de.qabel.core.http.DropServerHttp.DropServerResponse
 import de.qabel.core.repository.entities.DropState
 import org.slf4j.LoggerFactory
 
-class MainDropConnector(val dropServer: DropServerHttp) : DropConnector {
+class MainDropConnector(val dropServer: DropServerHttp):
+    DropConnector,
+    DropParser by DefaultDropParser() {
```
julianseeger: Does this have to be a delegate (is anyone using `parse()` from outside this class?) or can you just use a value?
```KOTLIN
val dropParser = DefaultDropParser()
...
dropParser.parse(...)
```
If so, I would prefer a value over a MainDropConnector with an unused parse method in its interface.

core/src/main/java/de/qabel/core/http/MainDropConnector.kt
```diff
@@ -31,36 +35,24 @@ class MainDropConnector(val dropServer: DropServerHttp) : DropConnector {
...
-                else -> {
-                    logger.warn("Received DropMessage with unknown binary version")
-                    null
-                }
-            }
+        val receivers = Identities().apply { put(identity) }
+        val messages = dropResult.third.map { byteMessage ->
             try {
-                binaryMessage?.disassembleMessage(identity)?.let { dropMessages.add(it) }
+                parse(byteMessage, receivers).second
```
julianseeger: What is `.first` for? I can only see one usage in a test that wouldn't have to test the identity explicitly.
If it's used anywhere else I would prefer an unpack (val (identity, message) = ...) instead of using .first .second and .third ... if you don't want to unpack, create a one-line-data-class so that you can do `parse(...).message` etc

julianseeger:
Code is much more readable now, +1 for this.
Three general things about using kotlin are do be discussed:
 - extensive use of `?.`
 - val over delegate if you don't need to implement an interface
 - using generic accessors of tuples

core/src/main/java/de/qabel/core/drop/DropSerializer.kt
JoGir: May use something like this instead of all this null-checks and exceptions
```
fun JsonObject.fGet(key: String): JsonElement = get(key) ?: throw JsonSyntaxException("Missing property " + key)

core/src/main/java/de/qabel/core/http/MainDropConnector.kt
JoGir: i dont think parse is used outside of the class, so a private val makes sense to hide the parser

core/src/main/java/de/qabel/core/http/MainDropConnector.kt
JoGir:
``` kotlin
val (statusCode, eTag, messages) = dropResult

PR #546, opened 2016-08-08
fix set Etag

### LINKED ISSUE #547: DropState not updated after retrieving drop messages
https://github.com/Qabel/qabel-core/blob/master/core/src/main/java/de/qabel/core/service/MainChatService.kt#L57

The current drop state is retrieved but after refreshing the messages, the new drop state is not set. This results in always pulling all messages.

JoGir:
Resolves #547


audax:
Resolves #547

PR #549, opened 2016-08-10
Send "eTag" in the correct header

audax:
Still missing the test because the MainDropServer doesn't have a test…


core/src/main/java/de/qabel/core/http/MainDropServer.kt
```diff
@@ -38,7 +38,7 @@ class MainDropServer : DropServerHttp {
     override fun receiveMessageBytes(uri: URI, eTag: String): Triple<Int, String, Collection<ByteArray>> =
         connect(uri).use {
             if (!eTag.isEmpty()) {
-                it.addRequestProperty(QblHeaders.X_QABEL_LATEST, eTag)
```
enkore: maybe don't call the property 'eTag', it just adds to the confusion

PR #554, opened 2016-08-12
Prepare contacts and chat

JoGir:
Required for https://github.com/Qabel/qabel-android/pull/622
* Fixes some bugs in contact <-> chat handling
* Adds paging to BaseRepository and implement for ChatDropMessages
* ContactData data class contains a contact with associated Identities
* Adds kotlin log interface for smart logging
* remove mutable/inconsistent entity fields from hashCode/equals methods


core/src/main/java/de/qabel/core/repository/framework/QueryBuilder.kt
```diff
@@ -143,11 +144,23 @@ class QueryBuilder {
...
         (if (!where.isEmpty()) where.toString() else "") + " " +
         (if (!groupBy.isEmpty()) groupBy.toString() else "") + " " +
+        (if (!orderBy.isEmpty()) orderBy.toString() else "") +
+        paging
+
+    fun countQueryString() : String = " SELECT count(*) " +
+        from.toString() + " " +
+        (if (!joins.isEmpty()) joins.toString() else "") + " " +
+        (if (!where.isEmpty()) where.toString() else "") + " " +
+        (if (!groupBy.isEmpty()) groupBy.toString() else "") + " " +
```
julianseeger: A lot of duplication here:
* queryString and countQueryString use exactly the same from, join, where groupBy and orderBy.
* join, where, groupBy and orderBy all repeat the pattern `(if (!collection.isEmpty()) collection.toString() else "")`

 Please extract at least the latter one to increase readability.
I could imagine a function like
```KOTLIN
listOf(
    select.toString(),
    from.toString,
    joinQuery,
    whereQuery,
    groupByQuery,
    orderByQuery
).join(" ")
```
If you even extract the first duplication, the intent (in this case the difference between the two methods) becomes very clear.

core/src/main/java/de/qabel/core/repository/sqlite/SqliteChatDropMessageRepository.kt
```diff
@@ -18,18 +19,28 @@ import de.qabel.core.repository.sqlite.schemas.ChatDropMessageDB.IDENTITY_ID
 import de.qabel.core.repository.sqlite.schemas.ChatDropMessageDB.PAYLOAD
 import de.qabel.core.repository.sqlite.schemas.ChatDropMessageDB.PAYLOAD_TYPE
 import de.qabel.core.repository.sqlite.schemas.ChatDropMessageDB.STATUS
+import de.qabel.core.repository.sqlite.schemas.ContactDB
 
 class SqliteChatDropMessageRepository(val database: ClientDatabase,
                                       entityManager: EntityManager) :
     BaseRepositoryImpl<ChatDropMessage>(ChatDropMessageDB, database, entityManager),
```
julianseeger: It's not from this diff but how did this happen? From the name `BaseRepositoryImpl` I would expect the `BaseRepository` class to be an abstract class that implements the `Repository` interface. So `Base` + `Repository` + `Impl` looks confusing doesn't it?

julianseeger:
Please link a ticket or mockup. I could use some context here ;)

mr-gosh:
I would upload one - but i don't get it either...

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
```diff
@@ -27,19 +29,19 @@ class SqliteContactRepository(db: ClientDatabase, em: EntityManager, dropUrlRepo
                               private val contactRelation: ContactDB = ContactDB(dropUrlRepository)) :
     BaseRepositoryImpl<Contact>(contactRelation, db, em), ContactRepository {
 
+    val log: Logger by lazy { LoggerFactory.getLogger(ContactRepository::class.java) }
```
julianseeger: `SqliteContactRepository`
Maybe it's time to build a more generic solution of this with
```KOTLIN
class LoggerAware {
    val log: Logger by Lazy { LoggerFactory.getLogger(this.javaClass) }
}

class SqliteContactRepository(...) : ..., LoggerAware {
...
```
If this works, then everyone could just extend the LoggerAware class and start logging with the correct name

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
```diff
@@ -125,34 +125,59 @@ class SqliteContactRepository(db: ClientDatabase, em: EntityManager, dropUrlRepo
...
 
-    override fun findWithIdentities(searchString: String): Collection<Pair<Contact, List<Identity>>> {
-        val identities = identityRepository.findAll()
+    override fun findWithIdentities(searchString: String, status: List<Contact.ContactStatus>, excludeIgnored: Boolean): Collection<Pair<Contact, List<Identity>>> {
+        log.info("findWithIdentities with filters {}, {}, {}", searchString, status.map { it.name }, excludeIgnored)
         val contacts = with(createEntityQuery()) {
+            //Exclude identities
+            leftJoin(ContactDB.IdentityJoin.TABLE, ContactDB.IdentityJoin.TABLE_ALIAS,
+                ContactDB.IdentityJoin.CONTACT_ID.exp(), relation.ID.exp())
+            whereAndNull(ContactDB.IdentityJoin.CONTACT_ID)
```
julianseeger: -> refactor -> extract method -> excludeIdentities

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
enkore: I would expect that `this.javaClass` would always become `LoggerAware`, but maybe there's a workaround there?

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
```diff
@@ -125,34 +125,59 @@ class SqliteContactRepository(db: ClientDatabase, em: EntityManager, dropUrlRepo
...
             getResultList<Contact>(this)
         }
+        val identities = identityRepository.findAll()
         val contactIdentities = findIdentityIds(contacts)
-
-        val resultList = mutableListOf<Pair<Contact, List<Identity>>>()
-        for (contact in contacts) {
-            resultList.add(Pair(contact, contactIdentities[contact.id]!!.map {
+        return contacts.map {
+            Pair(it, contactIdentities[it.id]!!.map {
```
julianseeger: not a TODO but ftr from the kotlin documentation:
> The standard library provides Pair and Triple. In most cases, though, named data classes are a better design choice, because they make the code more readable by providing meaningful names for properties.

core/src/main/java/de/qabel/core/repository/sqlite/schemas/ContactDB.kt
```diff
@@ -15,27 +15,39 @@ import java.sql.ResultSet
...
+        const val T_ALIAS = "c"
+        val ID = DBField("id", TABLE, T_ALIAS)
+        val ALIAS = DBField("alias", TABLE, T_ALIAS)
+        val PUBLIC_KEY = DBField("publicKey", TABLE, T_ALIAS)
+        val PHONE = DBField("phone", TABLE, T_ALIAS)
+        val EMAIL = DBField("email", TABLE, T_ALIAS)
+
+        val STATUS = DBField("status", TABLE, T_ALIAS)
+        val IGNORED = DBField("ignored", TABLE, T_ALIAS)
+        val NICKNAME = DBField("nickname", TABLE, T_ALIAS)
```
julianseeger:
```KOTLIN
internal fun field(name: String) = DBField(name, TABLE, T_ALIAS)
```

core/src/main/java/de/qabel/core/service/MainChatService.kt
```diff
@@ -47,16 +47,16 @@ open class MainChatService(val dropConnector: DropConnector, val identityReposit
...
+        val resultMap = DefaultHashMap<String, List<ChatDropMessage>>({ emptyList() })
         identityRepository.findAll().entities.forEach { identity ->
             identity.dropUrls.forEach { dropUrl ->
                 val dropState = dropStateRepository.getDropState(dropUrl)
                 logger.info("Fetching DropMessages from {} with eTag {}", dropState.drop, dropState.eTag)
                 try {
                     val dropResult = dropConnector.receiveDropMessages(identity, dropUrl, dropState)
                     val newMessages = handleDropUpdate(identity, dropResult.dropState, dropResult.dropMessages)
-                    resultMap.getOrDefault(identity).addAll(newMessages)
+                    resultMap.put(identity.keyIdentifier, resultMap.getOrDefault(identity.keyIdentifier).plus(newMessages))
```
julianseeger: I would prefer "fixing" the identities hashCode to be immutable. Then an error on this is impossible by design. Otherwise this problem will probably occur over and over again.

core/src/main/java/de/qabel/core/service/MainChatService.kt
```diff
@@ -47,16 +47,16 @@ open class MainChatService(val dropConnector: DropConnector, val identityReposit
         chatDropMessageRepository.update(message)
     }
 
-    override fun refreshMessages(): Map<Identity, List<ChatDropMessage>> {
-        val resultMap = DefaultHashMap<Identity, MutableList<ChatDropMessage>>({ mutableListOf() })
+    override fun refreshMessages(): Map<String, List<ChatDropMessage>> {
+        val resultMap = DefaultHashMap<String, List<ChatDropMessage>>({ emptyList() })
```
julianseeger: maybe this is TBD, but I would prefer keeping it a mutable list. The interface indicates an immutable list so for the user, nothing changes but we don't have to copy lists. Call it premature optimization but as it was a mutableList and because it doesn't matter for the design, why use an immutable list and mutate it then with a plus? I think it's always worse in case of performance and memory consumption ... even on this low scale, why changing an immutable list?

core/src/test/java/de/qabel/core/repository/SqliteChatDropMessageRepositoryTest.kt
```diff
@@ -20,9 +21,9 @@ class SqliteChatDropMessageRepositoryTest : AbstractSqliteRepositoryTest<ChatDro
     lateinit var identityRepo: IdentityRepository
     lateinit var contactRepo: ContactRepository
 
-    val identityA = IdentityBuilder(DropUrlGenerator("http://localhost")).withAlias("identityA").build()
-    val contactA = Contact("contactA", LinkedList<DropURL>(), QblECPublicKey("test13".toByteArray()));
-    val contactB = Contact("contactB", LinkedList<DropURL>(), QblECPublicKey("test24".toByteArray()));
+    val identityA : Identity = IdentityBuilder(DropUrlGenerator("http://localhost")).withAlias("identityA").build()
```
julianseeger: the IdentityBuilder didn't change. What's the problem with the type inference here?

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -215,25 +218,23 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
...
+        val contacts = repo.findWithIdentities()
 
-        val contacts = repo.findWithIdentities();
-        var index = 0;
         for (contact in contacts) {
-            val storedDto = storedContacts[index];
-            assertThat(storedDto.first.alias, equalTo(contact.first.alias));
-            assertThat(storedDto.second, hasSize(contact.second.size));
-            index++;
+            val storedDto = storedContacts.find { it.first.id == contact.first.id }!!
```
julianseeger: It's actually no longer the same test. The previous version was kind of obfuscated, but it asserted the order and partially the completenes of the `findWithIdentities()` result. That's no longer true.
It's fine to clean up a test but try not to remove implicit assertions but to make them explicit. For example if the result of contacts would contain two entries of `otherContact`, your test passes but the old one wouldn't. Additionally, you are implicitly testing the update methods ... better create a new test for them (or multiple) and test the permutations of it with seperate tooling.

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -242,23 +243,59 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
...
+        repo.save(contact, identity)
+        repo.save(unknownContact, identity)
+        repo.save(ignoredContact, identity)
+        repo.save(otherContact, identity)
+
+        //Use defaults
+        val contacts = repo.findWithIdentities()
+
+        Assert.assertEquals(2, contacts.size)
+        val fooContact = contacts.iterator().next()
```
julianseeger: contacts[0] ?

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -242,23 +243,59 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
...
+    fun testFindAllWithDefaults() {
+        repo.save(contact, identity)
+        repo.save(unknownContact, identity)
+        repo.save(ignoredContact, identity)
+        repo.save(otherContact, identity)
+
+        //Use defaults
+        val contacts = repo.findWithIdentities()
+
+        Assert.assertEquals(2, contacts.size)
```
julianseeger: import (static) and use the method only ..

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -242,23 +243,59 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
...
+        repo.save(contact, identity)
+        repo.save(unknownContact, identity)
+        repo.save(ignoredContact, identity)
+        repo.save(otherContact, identity)
+
+        val contacts = repo.findWithIdentities("", listOf(Contact.ContactStatus.NORMAL), false)
+        Assert.assertEquals(3, contacts.size)
+        val igContact = contacts.find { it.first.id == ignoredContact.id }!!
+        assertThat(ignoredContact.alias, equalTo(igContact.first.alias))
+        assertThat(1, equalTo(igContact.second.size))
```
julianseeger: That's the reason why using `Pair` is bad if you don't watch out what you are using it for.
`second` is probably a list of all connected identities but you can create a one-liner class and make this line
```KOTLIN
ignoredContact.identities shouldMatch hasSize(1)
```
without effort. Much more readable (btw.: abbreviate `ig`)

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -242,23 +243,59 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
...
+        val dropGen = DropUrlGenerator("http://mock.de")
+        val dropA = dropGen.generateUrl()
+        val dropB = dropGen.generateUrl()
+        contact.addDrop(dropA)
         repo.save(contact, identity)
+        contact.addDrop(dropB)
         contact.nickName = "testNick"
         repo.update(contact, listOf(identity, otherIdentity))
 
         val result = repo.findContactWithIdentities(contact.keyIdentifier)
```
julianseeger: use a custom class or unpack this to
`val (foundContact, foundIdentities) = ...` ... nobody want's to read code like `result.first`.
Keep in mind that the readability of testcode is even more important than that of the corresponding non-test code.

core/src/test/java/de/qabel/core/repository/inmemory/InMemoryChatDropMessageRepository.kt
```diff
@@ -5,9 +5,14 @@ import de.qabel.core.config.Identity
 import de.qabel.core.repository.ChatDropMessageRepository
 import de.qabel.core.repository.entities.ChatDropMessage
 import de.qabel.core.repository.exception.EntityNotFoundException
+import de.qabel.core.repository.framework.PagingResult
 
 class InMemoryChatDropMessageRepository : ChatDropMessageRepository {
 
+    override fun findByContact(contactId: Int, identityId: Int, offset: Int, pageSize: Int): PagingResult<ChatDropMessage> {
+        throw UnsupportedOperationException("Not implemented in Mock")
+    }
```
julianseeger: `= TODO()`

core/src/test/java/de/qabel/core/service/ChatServiceTest.kt
```diff
@@ -67,10 +67,10 @@ class ChatServiceTest {
 
         val result = chatService.refreshMessages()
         assertThat(result.keys, hasSize(1))
-        assertThat(result.keys.first().keyIdentifier, equalTo(identityB.keyIdentifier))
+        assertThat(result.keys.first(), equalTo(identityB.keyIdentifier))
 
         assertThat(result.values.first(), hasSize(1))
-        val received = result.values.first().first();
+        val received = result.values.first().first()
```
julianseeger: Please don't write testcode like this.
```KOTLIN
val messages = result.get(identityB.keyIdentifier)
messages shouldMatch hasSize(1)
val received = messages[0]
...
```
Now everyone knows what's happening here without guessing that result is a Map<Identity, List<Message>> and what you are doing with all these `first()` calls.

julianseeger:
> I would upload one - but i don't get it either...

Some kind of message history with a pagination including a total page count. But I can't find such an issue

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
audax: This can be an interface instead, see the AnkoLogger for example:
https://github.com/Kotlin/anko/blob/master/anko/library/static/common/src/Logger.kt

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
audax: or contacts.first()

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
audax: The data class can be even an inner class of the contact repository, or the data classes can be collected in a models.kt like in the index-server API.

audax:
Afaik this PR is for supporting ignore lists for contacts and pagination for chat messages.

@JoGir Next time, please put the purpose of your PR in the description along with the reason and, if applicable, connected PR in another repository.

julianseeger:
Ok now I've been informed that its for the RecyclerView and it's most easy to feed it with chunks of messages.
A new problem might arise from this PR: if we fetch for example messages 0 to 49 and we want to fetch messages 50 to 99 after that. But between the new fetches, a new message arrives. So the old message with the index 49 now has the index 50. And if we really fetch messages 50 to 99, then we have one duplicate. I don't know if android can handle this and automatically increments the maintained indexes when a new message arrives but audax had a nice idea for an interface where this cannot happen:
Fetch {amount of messages} since {oldest known message}.
```KOTLIN
fun getPagingResult(..., size: Int, since: Message)
```
adjust the name and type of `since: Message` as you wish but I hope this shows the idea. We don't have to know if the old index 49 is now 50 or maybe 51 because we just show the last known message and it's up to the repository to decide which messages follow that message.

core/src/main/java/de/qabel/core/repository/sqlite/SqliteContactRepository.kt
JoGir: Anko is for Android
```
Anko is a library which makes Android application development faster and easier.
It makes your code clean and easy to read, and lets you forget about rough edges of Android SDK for Java.

**Comment by JoGir** [comment]
resolves #566

File: core/src/main/java/de/qabel/core/repository/sqlite/migration/Migration1460997042ExtendContact.java [hunk 17]
```diff
@@ -19,7 +19,8 @@ public void up() throws SQLException {
         execute("ALTER TABLE contact ADD COLUMN nickname VARCHAR(255)");
         execute("ALTER TABLE contact ADD COLUMN status INTEGER NOT NULL DEFAULT 1");
         execute("ALTER TABLE contact ADD COLUMN ignored BOOLEAN NOT NULL DEFAULT FALSE");
-        execute("UPDATE contact set nickname=alias");
+        //Default STATUS=NORMAL and not ignored
+        execute("UPDATE contact set nickname=alias, ignored=false, status=1");
```
audax: Please put that into a new migration where you just set ignored=false and status=1 where ignored=NULL or status=NULL. Changing migrations is a bit nasty.

core/src/test/java/de/qabel/core/repository/inmemory/InMemoryContactRepository.kt
```diff
@@ -62,7 +62,10 @@ class InMemoryContactRepository : ContactRepository {
 
     override fun findContactWithIdentities(keyId: String): ContactData {
         if (contacts.contains(keyId)) {
```
audax: Didn't see that before, but we can fix this now:
`contacts[keyId]?.let { contact -> ...}`
and you don't have to use !! anymore.

JoGir:
resolves #566

PR #556, opened 2016-08-16
save throws Exception if contact already exists

nnice:


core/src/test/java/de/qabel/core/repository/inmemory/InMemoryContactRepository.kt
```diff
@@ -27,7 +27,11 @@ class InMemoryContactRepository : ContactRepository {
     }
 
     override fun save(contact: Contact, identity: Identity) {
-        contact.id = contacts.size + 1
+        if (contact.id == 0 && exists(contact)) {
+            throw EntityNotFoundException("Contact already exists!")
```
enkore: That looks strange. "Entity not found - Already exists" — add new exception?

nnice: yeah sure, didnt knew it existed already. gonna change it

core/src/test/java/de/qabel/core/repository/InMemoryContactRepositoryTest.kt
```diff
@@ -18,8 +20,8 @@ class InMemoryContactRepositoryTest {
         val identityKey = QblECKeyPair()
```
julianseeger: pull this out as a class property. Then you don't need to redeclare it in every test

PR #561, opened 2016-08-18
EntityNotFound exception added in delete

nnice:


julianseeger:
usecase?

nnice:
Its used in Desktop.  Good mention  from @jogir  to add this and for save  exception into the  real  impl for  sqliterepo

julianseeger:
Ok, I actually don't understand the use case. It's for deleting contacts that don't exist ... shouldn't happen anyways (only as a race condition between the gui and a cross-client-sync or similar) and if so skipping it would be an option, too. Then nobody would need that in a TestDouble ?! Am I missing something?

julianseeger:
FTR: the resolution is to remove the EntityNotFoundException from the real repository, then the InMemoryRepository doesn't need it either.
It even complies with the block-server: if you wish to delete it and it's already deleted, then everything is fine and we take it as a success.

PR #568, opened 2016-08-25
Merge identical files when updating

### LINKED ISSUE #492: Conflict By Same Name
Folder a/b/c → 	…	← a/b/c
create data clutter

Solution:
File hashes into dm 
size & mtime for prediction

audax:
When a file is updated and the existing file
has the same hash, the update is ignored
and the new block is deleted

Fixes #492

### DIFF
--- a/box/src/main/java/de/qabel/box/storage/command/UpdateFileChange.kt
+++ b/box/src/main/java/de/qabel/box/storage/command/UpdateFileChange.kt
@@ -1,13 +1,13 @@
 package de.qabel.box.storage.command
 
-import de.qabel.box.storage.BoxFile
-import de.qabel.box.storage.BoxFolder
-import de.qabel.box.storage.BoxObject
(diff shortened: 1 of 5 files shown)


box/src/main/java/de/qabel/box/storage/command/UpdateFileChange.kt
```diff
@@ -21,6 +21,10 @@ open class UpdateFileChange(val expectedFile: BoxFile?, private val newFile: Box
             dm.insertFile(newFile)
         } catch (e: QblStorageNameConflict) {
             with (findCurrentFileOrFolder(dm, filename)) {
+                if (this is BoxFile && this.isHashed() && this.hashed == newFile.hashed) {
+                    writeBackend.deleteBlock(newFile.block)
```
julianseeger: This looks bad. The execute method may be executed multiple times with different input.
There is a postprocess method for undoable actions like deleting a block.

PR #571, opened 2016-08-29
Force commit of index navigation after sharing

### LINKED ISSUE #543: sharing sender doesn't keep the shares
the indexDM is currently not handled correctly. For example when sharing a file (from a submfolder), the indexDM is changed but not uploaded => the share that has been added to the list of shares in the indexDM is never uploaded.

audax:
Autocommit mitigated most of the problem, but only if
another change like creating a folder happened. Just sharing a
file from the subfolder did not force a commit in the index navigation
Fixes #543

### DIFF
--- a/box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
+++ b/box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
@@ -708,6 +708,22 @@ public abstract class BoxVolumeTest {
         assertThat(subnavB2.listFiles(), hasSize(1));
     }
 
+    @Test
+    public void shareInsertedInIndexNavigationWhenSharingFromFolder() throws Exception {
+        BoxNavigation nav = volume.navigate();
+        nav.setAutocommit(false);
+        BoxFolder folder = nav.createFolder("folder");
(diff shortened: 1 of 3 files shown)


box/src/test/java/de/qabel/box/storage/BoxVolumeTest.java
```diff
@@ -708,6 +708,22 @@ public void conflictsBySameNameInSubfolders() throws Exception {
...
+    @Test
+    public void shareInsertedInIndexNavigationWhenSharingFromFolder() throws Exception {
+        BoxNavigation nav = volume.navigate();
+        nav.setAutocommit(false);
+        BoxFolder folder = nav.createFolder("folder");
+        BoxNavigation subNav = nav.navigate(folder);
+        File file = new File(testFileName);
+        BoxFile boxFile = subNav.upload("file1", file);
+        subNav.share(keyPair.getPub(), boxFile, contact.getKeyIdentifier());
+        subNav.commit();
```
julianseeger: the test would be more interesting if you switched line 719 and 720, wouldn't it? :D

PR #573, opened 2016-08-29
contact repository observable

taynaps:
#387 implemented contact repository observable

### DIFF
--- a/core/src/main/java/de/qabel/core/config/ContactObserver.java
+++ b/core/src/main/java/de/qabel/core/config/ContactObserver.java
@@ -1,5 +1,5 @@
 package de.qabel.core.config;
 
-interface EntityObserver {
+public interface ContactObserver {
     void update();
 }
--- a/core/src/main/java/de/qabel/core/config/ContactObserver.java
+++ b/core/src/main/java/de/qabel/core/config/ContactObserver.java
@@ -1,5 +1,0 @@
-package de.qabel.core.config;
-
-public interface ContactObserver {
-    void update();
-}
--- a/core/src/main/java/de/qabel/core/config/EntityDelegate.kt
+++ b/core/src/main/java/de/qabel/core/config/EntityDelegate.kt
@@ -2,11 +2,11 @@ package de.qabel.core.config
 
 class EntityDelegate : EntityObservable {
 
-    override fun removeObserver(observer: EntityObserver) {
-        throw UnsupportedOperationException("not implemented")
-    }
-
     private val observers: MutableList<EntityObserver> = mutableListOf()
+
+    override fun removeObserver(observer: EntityObserver) {
+        observers.remove(observer);
+    }
 
     override fun attach(observer: EntityObserver) {
         observers += observer
--- a/core/src/main/java/de/qabel/core/config/EntityDelegate.kt
+++ b/core/src/main/java/de/qabel/core/config/EntityDelegate.kt
@@ -1,0 +1,16 @@
+package de.qabel.core.config
+
+class EntityDelegate : EntityObservable {
+
+    override fun removeObserver(observer: EntityObserver) {
+        throw UnsupportedOperationException("not implemented")
+    }
+
+    private val observers: MutableList<EntityObserver> = mutableListOf()
+
+    override fun attach(observer: EntityObserver) {
+        observers += observer
+    }
+
+    override fun notifyObservers() = observers.forEach { it.update() }
+}
--- a/core/src/main/java/de/qabel/core/config/IdentityObserver.java
+++ b/core/src/main/java/de/qabel/core/config/IdentityObserver.java
@@ -1,6 +1,6 @@
 package de.qabel.core.config;
 
-public interface IdentityObserver {
-    public abstract void update();
+interface IdentityObserver {
+    void update();
 }
 
--- a/core/src/test/java/de/qabel/core/repository/InMemoryContactRepositoryTest.kt
+++ b/core/src/test/java/de/qabel/core/repository/InMemoryContactRepositoryTest.kt
@@ -10,12 +10,11 @@ import de.qabel.core.repository.inmemory.InMemoryContactRepository
 import org.hamcrest.Matchers.hasSize
 import org.junit.Test
 import org.junit.Assert.*
-import java.util.concurrent.atomic.AtomicBoolean
 
 class InMemoryContactRepositoryTest {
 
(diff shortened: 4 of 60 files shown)


core/src/main/java/de/qabel/core/config/ContactObserver.java
```diff
@@ -0,0 +1,5 @@
+package de.qabel.core.config;
+
+public interface ContactObserver {
+    void update();
```
julianseeger: Why do you need so many observers if they all have the same interface?
```JAVA
    void update();
```

You should either stick to one common interface or give that diversity a meaning by
a) using it for the updated instance `void update(Contact updatedContact)`
b) use it to ease the observer handling `observers.filter { it is ContactObserver }.forEach { it.update() }`
c) both

core/src/test/java/de/qabel/core/repository/InMemoryContactRepositoryTest.kt
```diff
@@ -52,4 +44,24 @@ class InMemoryContactRepositoryTest {
...
+    private fun attachContactObserver() {
+        repo.attach(ContactObserver { hasCalled.set(true) })
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun testContactRepositorySaveObservable() {
+        attachContactObserver()
+        repo.save(contact, identity)
+        Assert.assertTrue(hasCalled.get())
```
julianseeger: `assertTrue(hasCalled.get())` is possible in kotlin, too.
Just `import org.junit.Assert.assertTrue`

core/src/test/java/de/qabel/core/repository/inmemory/InMemoryContactRepository.kt
```diff
@@ -95,4 +99,14 @@ open class InMemoryContactRepository : ContactRepository {
             identities.put(it.keyIdentifier, it)
         }
     }
+
+    fun attach(observer: ContactObserver) {
```
julianseeger: to apply some kind of collective code ownership, try to use the short notation for method bodies if they consist of only one line:

`fun attach(observer: ContactObserver) = observers.add(observer)`
that should work here, is much shorter and tells the reader that there is no more logic than that

core/src/test/java/de/qabel/core/repository/inmemory/InMemoryContactRepository.kt
```diff
@@ -95,4 +99,14 @@ open class InMemoryContactRepository : ContactRepository {
...
+        observers.add(observer)
+    }
+
+    fun notifyAllObservers() {
```
julianseeger: fun notifyAllObservers = observers.forEach { it.update() }

julianseeger:
What about the SqliteContactRepository? If we want to use these Observers, the SqliteContactRepository is probably the most important one

core/src/main/java/de/qabel/core/config/EntityDelegate.kt
```diff
@@ -0,0 +1,16 @@
+package de.qabel.core.config
+
+class EntityDelegate : EntityObservable {
+
+    override fun removeObserver(observer: EntityObserver) {
+        throw UnsupportedOperationException("not implemented")
```
audax: still not implemented. If it is not yet needed, you may just remove the method from the interface and class.

core/src/test/java/de/qabel/core/repository/InMemoryContactRepositoryTest.kt
```diff
@@ -52,4 +43,24 @@ class InMemoryContactRepositoryTest {
     private fun createIdentity(identityName: String, identityKey: QblECKeyPair): Identity {
         return Identity(identityName, emptyList(), identityKey)
     }
+
+    private fun attachEntityObserver() {
+        repo.attach(EntityObserver { hasCalled.set(true) })
+    }
+
+    @Test
+    @Throws(Exception::class)
```
audax: no need for this annotation in pure kotlin code. Those `@Throws` annotation are purely for Java inter-op

core/src/test/java/de/qabel/core/repository/SqliteContactRepositoryTest.kt
```diff
@@ -28,6 +30,7 @@ class SqliteContactRepositoryTest : AbstractSqliteRepositoryTest<SqliteContactRe
     private lateinit var pubKey: QblECPublicKey
     private lateinit var identityRepository: SqliteIdentityRepository
     private lateinit var dropUrlGenerator: DropUrlGenerator
+    private val hasCalled = AtomicBoolean()
```
audax: You can use a normal boolean here, this is not multithreaded

core/src/main/java/de/qabel/core/config/EntityDelegate.kt
```diff
@@ -0,0 +1,16 @@
+package de.qabel.core.config
+
+class EntityDelegate : EntityObservable {
```
julianseeger: If you only see `EntityDelegate` - which a user might do - one might think that its a delegate for an Entity or something that delegates to an entity. Thus, this is not a good name.
If you want to say that it's an EntityObservable and the specific thing about it is that it's a delegate, then `DelegateEntityObservable` would be a typical naming scheme.
But looking at the class, there is nothing delegate specific, it's just a pure implementation of an EntityObservable, so you might choose a name for that like `SimpleEntityObservable`, `PureEntityObservable` or - what I personally don't like but a lot of developers would do - `EntityObservableImpl`. The important thing here is that one can take the SimpleName of that class and have a quick overview of what the class does / represents. Typically more specific from right to left.

core/src/main/java/de/qabel/core/config/IdentityObserver.java
```diff
@@ -1,6 +1,6 @@
 package de.qabel.core.config;
 
-public interface IdentityObserver {
-    public abstract void update();
+interface IdentityObserver {
```
julianseeger: the question still holds: what is `Identity` or `Entity` specific in this interfaces?
If you don't pass what changed (`void update(identity: Identity)`) then where is the need for multiple Interfaces? You could create one `Observer` and one `Observable` interface and use it instead of all your specific interfaces. You wouldn't loose anything but superfluous code imo.

audax: I think that diff only exists to remove public and abstract modifiers and is not meant as the final implementation

PR #576, opened 2016-08-29
Share - Chat integration

JoGir:
* extends chatservice with filesharing
* add handling for receiving shares

dependency of https://github.com/Qabel/qabel-android/issues/562

### DIFF
--- a/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
+++ b/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
@@ -337,19 +337,15 @@ abstract class AbstractNavigation(
 
     @Throws(QblStorageException::class, IOException::class, InvalidKeyException::class)
     override fun updateFileMetadata(boxFile: BoxFile) {
-        val meta = boxFile.meta
-        val metakey = boxFile.metakey
-        if (meta == null || metakey == null) {
-            throw IllegalArgumentException("BoxFile without FileMetadata cannot be updated")
-        }
+        val shared = boxFile.shared ?: throw IllegalArgumentException("BoxFile without FileMetadata cannot be updated")
+
         try {
-            val out = getMetadataFile(meta, metakey)
-            val fileMetadataOld = fileFactory.open(out)
+            val fileMetadataOld = getMetadataFile(shared)
             val fileMetadataNew = fileFactory.create(
                 fileMetadataOld.file?.owner ?: throw QblStorageException("No owner in old file metadata"),
                 boxFile
             )
-            uploadEncrypted(fileMetadataNew.path, KeyParameter(metakey), meta)
+            uploadEncrypted(fileMetadataNew.path, KeyParameter(shared.metaKey), shared.meta)
         } catch (e: QblStorageException) {
             logger.error("Could not create or upload FileMetadata", e)
             throw e
@@ -362,15 +358,10 @@ abstract class AbstractNavigation(
 
     @Throws(IOException::class, InvalidKeyException::class, QblStorageException::class)
     override fun getFileMetadata(boxFile: BoxFile): FileMetadata {
-        val meta = boxFile.meta
-        val metakey = boxFile.metakey
-        if (meta == null || metakey == null) {
-            throw IllegalArgumentException("BoxFile without FileMetadata cannot be updated")
-        }
+        val shared = boxFile.shared ?: throw IllegalArgumentException("BoxFile without FileMetadata cannot be loaded")
 
         try {
-            val out = getMetadataFile(meta, metakey)
-            return fileFactory.open(out)
+            return getMetadataFile(shared)
         } catch (e: QblStorageException) {
             logger.error("Could not create or upload FileMetadata", e)
             throw e
@@ -382,14 +373,12 @@ abstract class AbstractNavigation(
     }
 
     @Throws(QblStorageException::class, IOException::class, InvalidKeyException::class)
-    private fun getMetadataFile(meta: String, key: ByteArray): File {
-        readBackend.download(meta).inputStream.use { encryptedMetadata ->
-
+    override fun getMetadataFile(share: Share): FileMetadata {
+        readBackend.download(share.meta).inputStream.use { encryptedMetadata ->
             val tmp = File.createTempFile("dir", "db1", tempDir)
             tmp.deleteOnExit()
-            val keyParameter = KeyParameter(key)
-            if (cryptoUtils.decryptFileAuthenticatedSymmetricAndValidateTag(encryptedMetadata, tmp, keyParameter)) {
-                return tmp
+            if (cryptoUtils.decryptFileAuthenticatedSymmetricAndValidateTag(encryptedMetadata, tmp, KeyParameter(share.metaKey))) {
+                return fileFactory.open(tmp)
             } else {
                 throw QblStorageNotFound("Invalid key")
             }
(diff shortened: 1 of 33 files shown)


box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -362,15 +358,10 @@ abstract class AbstractNavigation(
 
     @Throws(IOException::class, InvalidKeyException::class, QblStorageException::class)
     override fun getFileMetadata(boxFile: BoxFile): FileMetadata {
-        val meta = boxFile.meta
-        val metakey = boxFile.metakey
-        if (meta == null || metakey == null) {
-            throw IllegalArgumentException("BoxFile without FileMetadata cannot be updated")
-        }
+        val shared = boxFile.shared ?: throw IllegalArgumentException("BoxFile without FileMetadata cannot be loaded")
```
audax: Throw a subclass of QblStorageException instead

chat/src/main/java/de/qabel/chat/repository/ChatShareRepository.kt
```diff
@@ -0,0 +1,22 @@
...
+import de.qabel.core.config.Identity
+import de.qabel.core.repository.framework.Repository
+
+interface ChatShareRepository : Repository<BoxFileChatShare> {
+
+    fun findByBoxReference(identity: Identity, metaUrl : String, metaKey : ByteArray): BoxFileChatShare?
+    fun find(identity: Identity, contact: Contact? = null) : List<BoxFileChatShare>
+
+    fun connectWithMessage(chatDropMessage: ChatDropMessage, share: BoxFileChatShare)
+    fun findShareChatDropMessageIds(share: BoxFileChatShare): List<Int>
```
julianseeger: If you only need this in tests, don't expose ids in the interface. you can expose this method to the tested repository and maybe even keep it package local

chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private fun tryRefreshShare(share: BoxFileChatShare, boxNavigation: BoxNavigation): BoxExternalFile? {
+        val boxExternalFile = try {
+            val fileMetadata = boxNavigation.getMetadataFile(Share(share.metaUrl, share.metaKey.byteList.toByteArray()))
+            val fileRef = fileMetadata.file!! //TODO
+            applyBoxFile(share, fileRef)
+            fileRef
+        } catch (deleted: QblStorageNotFound) {
+            share.status = ShareStatus.DELETED
+            null
```
julianseeger: Do or do not. There is no try.
[just throw the exception here because returning null is bad as we know]

chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
```diff
@@ -0,0 +1,128 @@
...
+import java.net.URISyntaxException
+import java.security.InvalidKeyException
+
+class MainSharingService(private val chatShareRepository: ChatShareRepository,
+                         private val contactRepository: ContactRepository,
+                         private val boxReadBackend: StorageReadBackend,
+                         private val cryptoUtils: CryptoUtils = CryptoUtils()) : SharingService {
+
+    override fun getOrCreateFileShare(identity: Identity, contact: Contact,
+                                      boxFile: BoxFile, boxNavigation: BoxNavigation): BoxFileChatShare =
```
julianseeger: this is not a one-liner. don't try to make one out of it

chat/src/main/java/de/qabel/chat/service/SharingService.kt
```diff
@@ -0,0 +1,27 @@
...
+import de.qabel.core.config.Identity
+import java.io.File
+
+interface SharingService {
+
+    fun getOrCreateFileShare(identity: Identity, contact: Contact, boxFile: BoxFile,
+                             boxNavigation: BoxNavigation): BoxFileChatShare
+
+    fun addMessageToShare(share: BoxFileChatShare, chatDropMessage: ChatDropMessage)
+    fun markShareSent(share: BoxFileChatShare): BoxFileChatShare
```
julianseeger: marking a share sent sounds like an internal task. why should the service care about that? The service looks like a facade and thus shouldn't expose such internal tasks.

chat/src/main/java/de/qabel/chat/service/SharingService.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun revokeFileShare(contact: Contact, share: BoxFileChatShare,
+                        boxFile: BoxFile, boxNavigation: BoxNavigation)
```
julianseeger: That interface implies that it really revokes anything. But it actually deletes the share and I don't know why we need a contact for that.

chat/src/main/java/de/qabel/chat/service/SharingService.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    fun refreshShare(share: BoxFileChatShare, boxNavigation: BoxNavigation): BoxExternalFile
+
+    fun receiveShare(identity: Identity, message: ChatDropMessage, payload: ChatDropMessage.MessagePayload.ShareMessage)
```
julianseeger: the message contains it's payload, doesn't it?
And at least I don't know what receiving a share means. I would appreceate a comment or another name.

chat/src/main/java/de/qabel/chat/service/SharingService.kt
JoGir: it actually unshares the file and updates the chatShare Entry. the contact currently isn required, out of scope... I´ll remove it...

chat/build.gradle
```diff
@@ -13,10 +13,13 @@ repositories {
     mavenCentral()
 }
 
+compileTestJava.dependsOn tasks.getByPath(':core:testClasses')
 dependencies {
     compile project(':core')
     compile project(':box')
-    //TODO bad include of test sources
+    // First dependency is for IDEA
+    testCompile files { project(':core').sourceSets.test.java.srcDirs }
```
julianseeger: You can disable the option "Create separate module per source set" [at the gradle properties](http://stackoverflow.com/questions/36372571/intellij-idea-and-gradle-why-there-are-3-modules-per-sub-module).
That should solve the problem as the test sources are imported with the project dependency.
That's how it works with the box module and it only needs the `compile project(':core')`

core/src/main/java/de/qabel/core/repository/framework/BaseRepository.kt
```diff
@@ -125,6 +131,7 @@ abstract class BaseRepository<T : BaseEntity>(val relation: DBRelation<T>,
                 })
             })
         } catch (e: SQLException) {
+            e.printStackTrace()
```
audax: booohhh

audax:
@JoGir This PR wants to merge into Qabel:develop, not master!!

chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
JoGir: Nullable is not always evil, especially when using kotlin. All nullable usages are private and external usages are not null...

audax: at least the fileRef still has a TODO

audax:
I didn't even know that I can change the base branch in a PR...

chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
julianseeger: > All nullable usages are private and external usages are not null...

Then let me rephrase: why don't you inline that method? Don't see why you need that indirection and that null conversion. And you could even throw exceptions with more explicit messages (like definitely permanently not reachable vs. maybe only currently not reachable)

audax:
For the share-status extension, see here:
https://github.com/Qabel/qabel-android/issues/644

chat/build.gradle
JoGir: not working for test sources

enkore:
> I didn't even know that I can change the base branch in a PR...

Neither did I until recently! :) The edit button near the headline (not the one on the 1st post) allows this.

PR #584, opened 2016-08-31
Fix initials method for contacts and identities

JoGir:


core/src/main/java/de/qabel/core/ui/EntityUIExtensions.kt
```diff
@@ -5,9 +5,11 @@ import de.qabel.core.config.Entity
...
 
 private val splitRegex: Regex by lazy { " ".toRegex() }
-private fun String.toInitials(): String = split(splitRegex).let {
-    first().toUpperCase() + if (it.size > 1) it[1].toUpperCase() else ""
-}
+private fun String.toInitials(): String = split(splitRegex).mapIndexed { i, s ->
+    if (i < 2 && s.length > 0) {
+        s.first().toUpperCase()
+    } else null
+}.filterNotNull().joinToString("")
```
audax: `split(splitRegex).take(2).map{ it.take(1).toUpperCase() }.joinToString("")`

PR #586, opened 2016-09-01
assertThrows with even more steroids

enkore:


enkore:
einmal durchwinken pls

PR #587, opened 2016-09-03
Chatservice tests

JoGir:
some extra tests for chatservice

### DIFF
--- a/chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
+++ b/chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
@@ -3,6 +3,7 @@ package de.qabel.chat.service
 import de.qabel.box.storage.*
 import de.qabel.box.storage.exceptions.QblStorageException
 import de.qabel.box.storage.exceptions.QblStorageNotFound
+import de.qabel.box.storage.jdbc.JdbcFileMetadataFactory
 import de.qabel.chat.repository.ChatShareRepository
 import de.qabel.chat.repository.entities.BoxFileChatShare
 import de.qabel.chat.repository.entities.ChatDropMessage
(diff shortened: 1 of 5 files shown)


chat/src/test/java/de/qabel/chat/service/ChatServiceTest.kt
```diff
@@ -203,4 +207,37 @@ class ChatServiceTest : CoreTestCase {
...
+        val volumeA = BoxVolumeImpl(LocalReadBackend(tempFolder.toPath()), LocalWriteBackend(tempFolder.toPath()),
+            identityA.primaryKeyPair, CryptoUtils().getRandomBytes(16), tempFolder, "").apply {
+            createIndex("qabel", "test123")
+        }
+        val navigationA = volumeA.navigate()
+        val testFile = randomFile(100)
+        val boxFile = navigationA.upload("TestFile", testFile)
+
+        //This is required, because sharing uses the contact_id of an identity as ownerContactId
+        contactARepository.save(contactA.apply { id == 0 }, identityA)
```
audax: I don't think you intented to use `==` here. This does not set the id to 0.

chat/src/test/java/de/qabel/chat/service/ChatServiceTest.kt
```diff
@@ -203,4 +207,37 @@ class ChatServiceTest : CoreTestCase {
...
+        val payload = result.payload as ChatDropMessage.MessagePayload.TextMessage
+        assertThat(payload.msg, equalTo("Hello"))
+        assertThat(result.contactId, equalTo(contactB.id))
+        assertThat(result.identityId, equalTo(identityA.id))
+    }
+
+    @Test
+    fun testSendShareMsg() {
+        val tempFolder = createTempDir("share_test")
+        val volumeA = BoxVolumeImpl(LocalReadBackend(tempFolder.toPath()), LocalWriteBackend(tempFolder.toPath()),
```
julianseeger: This doesn't add info to the test, so you better move it out of the test body (maybe a lazy property?). The important information here is that a file has been uploaded and the receiver of the share received that file. So everything until line 228 is not relevant to understand the test.

julianseeger:
what's the state here? can this be merged or does it get more tests?

JoGir:
moved readbackend to method params caused by identity dependencies

JoGir:
need fast merge for integrating this and the logging pr in android

chat/src/main/java/de/qabel/chat/service/MainSharingService.kt
```diff
@@ -122,4 +121,13 @@ class MainSharingService(private val chatShareRepository: ChatShareRepository,
...
             boxFile.block, contactRepository.findByKeyId(identity.keyIdentifier).id,
             identity.id)
+
+    private fun downloadFileMetadata(share: BoxFileChatShare, boxReadBackend: StorageReadBackend): BoxExternalFile {
+        val tmpFile = Files.createTempFile(tmpDir.toPath(), "tmp_", "_fm").toFile()
+        boxReadBackend.download(share.metaUrl, null).use({ download ->
+            cryptoUtils.decryptFileAuthenticatedSymmetricAndValidateTag(download.inputStream,
+                tmpFile, KeyParameter(share.metaKey.toByteArray()))
+        })
+        return JdbcFileMetadataFactory(tmpDir).open(tmpFile).file
```
julianseeger: fyi: we should start moving those factories out and inject them. Now that they are no longer static calls, we can do that and no longer pass those tmpDirs through all classes.

PR #588, opened 2016-09-03
Adaptive logging for all platforms

### LINKED ISSUE #567: Logging fail on android

JoGir:
resolves #567


julianseeger:
+1 for this. But I would prefer `Qabel` over `Qbl`. So either `QabelLogger` or just the `Logger` in the `de.qabel.*` package.

audax:
+1 for QabelLogger and not just Logger.

julianseeger:
> This branch has conflicts that must be resolved

julianseeger:
Out of interest: do we have any tests for this (or the old version)? Can't find any...

core/src/main/java/de/qabel/core/logging/DefaultQabelLogger.kt
```diff
@@ -0,0 +1,28 @@
+package de.qabel.core.logging
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+
+class DefaultQabelLogger(clazz: Class<*>) : QabelLoggerWrapper {
+
+    val logger: Logger = LoggerFactory.getLogger(clazz)
```
julianseeger: by lazy. we don't need to initialize a logger until someone actually logs. especially on android where those classes might get garbage collected before they are not needed ^^

core/src/main/java/de/qabel/core/logging/QabelLogger.kt
```diff
@@ -0,0 +1,21 @@
+package de.qabel.core.logging
+
+interface QabelLogger {
+    val logger: QabelLoggerWrapper
+        get() = QabelLoggerManager.factory.createLogger(javaClass)
```
julianseeger: Pretty sure you are creating a new DefaultQabelLogger instance for each log message. If you make this val logger lazy instead of defining a getter, we should get one instance per QabelLogger implementation, which should be fine.

core/src/main/java/de/qabel/core/logging/QabelLogger.kt
```diff
@@ -0,0 +1,21 @@
...
+}
+
+fun QabelLogger.trace(message: Any, vararg args: Any) =
```
julianseeger: why are these extension methods? I would prefer them inside the interface body or doesn't Kotlin allow default implementations?

core/src/main/java/de/qabel/core/logging/QabelLoggerWrapper.kt
```diff
@@ -0,0 +1,12 @@
+package de.qabel.core.logging
+
+interface QabelLoggerWrapper {
```
julianseeger: The naming goes the wrong way here. The QabelLogger looks like an implementation of the QabelLoggerWrapper

core/src/main/java/de/qabel/core/logging/QabelLogger.kt
JoGir: the using class only imports required logging methods...

core/src/main/java/de/qabel/core/logging/QabelLogger.kt
JoGir: Delegated properties not allowed in interfaces. looking for solutions...

PR #589, opened 2016-09-05
extending EntityObservable

taynaps:
#573


julianseeger:
did it work with the desktop client? (to delegate the inherited interface and attach using the ContactRepository interface?) then I would merge...

taynaps:
It worked in the desktop client.

PR #591, opened 2016-09-07
IndexServer.confirmVerificationCode, .denyVerificationCode

### LINKED ISSUE #590: index: API wrapper for confirmations
qabel/qabel.github.io#182

Really small thing, but should be in core to avoid unnecessary duplication.

enkore:
Fixes #590

Tiny bit untested, but should work.

### DIFF
--- a/core/src/main/java/de/qabel/core/index/IndexServer.kt
+++ b/core/src/main/java/de/qabel/core/index/IndexServer.kt
@@ -55,9 +55,10 @@ interface IndexServer {
     /**
      * Confirm a verification code.
      *
-     * Additional exceptions: [CodeInvalidException], [CodeExpiredException]
+     * These are sent to the user to verify eg. mail addresses or phone numbers.
+     * See [Index spec](http://qabel.github.io/docs/Qabel-Index/#verification-confirmation) for details.
      */
-    @Throws(IOException::class)
+    @Throws(IOException::class, CodeInvalidException::class, CodeExpiredException::class)
     fun confirmVerificationCode(code: String)
 
     /**
@@ -65,6 +66,6 @@ interface IndexServer {
      *
      * See [confirmVerificationCode]
      */
-    @Throws(IOException::class)
+    @Throws(IOException::class, CodeInvalidException::class, CodeExpiredException::class)
     fun denyVerificationCode(code: String)
 }
--- a/core/src/main/java/de/qabel/core/index/IndexServer.kt
+++ b/core/src/main/java/de/qabel/core/index/IndexServer.kt
@@ -51,4 +51,20 @@ interface IndexServer {
      */
     @Throws(IOException::class)
     fun updateIdentity(identity: UpdateIdentity): UpdateResult
+
+    /**
+     * Confirm a verification code.
+     *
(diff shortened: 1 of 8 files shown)


core/src/main/java/de/qabel/core/index/IndexServer.kt
```diff
@@ -51,4 +51,20 @@ interface IndexServer {
      */
     @Throws(IOException::class)
     fun updateIdentity(identity: UpdateIdentity): UpdateResult
+
+    /**
+     * Confirm a verification code.
+     *
+     * Additional exceptions: [CodeInvalidException], [CodeExpiredException]
```
julianseeger: Can't you add them to the @Throws annotation?

core/src/test/java/de/qabel/core/index/IndexHTTPTest.kt
```diff
@@ -141,4 +139,17 @@ class IndexHTTPTest {
         }
     }
 
+    @Test
+    fun testConfirmCodeInvalid() {
+        assertThrows(CodeInvalidException::class) {
```
julianseeger: in this scenario you could just use
```KOTLIN
@Test(expected=CodeInvalidException::class)
fun testConfirmCodeInvalid() = index.confirmVerificationCode("12345")
```

assertThrows simplifies especially those tests where multiple parts of the testcode could throw that exception but you want to make sure that a specific part throws it

core/src/test/java/de/qabel/core/index/VerificationCodeEndpointTest.kt
```diff
@@ -0,0 +1,16 @@
...
+
+
+class VerificationCodeEndpointTest {
+    private val server = IndexHTTPLocation("http://localhost:9698")
+    private val verification = VerificationCodeEndpointImpl(server)
+
+    @Test
+    fun testBuildRequestMethod() {
+        val request = verification.buildRequest("1234", confirm=true)
+        assertEquals("GET", request.method)
```
julianseeger: imho you should either assert the url, too or just delete the test if it only asserts the last line of that method.

core/src/main/java/de/qabel/core/index/IndexServer.kt
enkore: They are \<IndexServerException \< IOException, so they are already included. I just mention them here so they are easier to find. Or should that also go into the @Throws?

julianseeger: I see no reason why not. It's important enough to add it to the comment, so adding it to throws will tell the reader the same .. and you save that (otherwise) useless comment.

enkore: Ok

julianseeger:
Any chance to set up a scenario on the local index server to actually execute the get request and test if the server really responds on that url ?

enkore:
Yes, that's possible, but would require changes to the other tests as well.

I'm thinking along the lines of using Django's file-based mail backend, which would allow the tests to "read" the verification email and test that as well. Not sure how that would work with containers, though.

Another time?

julianseeger:
+

PR #592, opened 2016-09-07
Prepare index integration

JoGir:
Adds interactor, prepares for android integration
required for https://github.com/Qabel/qabel-android/issues/484 and https://github.com/Qabel/qabel-android/issues/474

- [x] base functionality for uploading identities
- [x] base functionality for syncing contacts
- [x] add and integrate verification flags in identity
- [x] normalize/formatting and validation extension for phone numbers 
- [x] search method for mail and phone for https://github.com/Qabel/qabel-desktop/issues/412 and https://github.com/Qabel/qabel-android/issues/648


core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
```diff
@@ -0,0 +1,226 @@
...
+            indexServer.updateIdentity(it)
+        }
+    }
+
+    /**
+     * Search the index for the external contacts and updates the local contact repository.
+     * Returns a list of new contacts.
+     *
+     * TODO With my last android researches my phone would requires 265 HttpRequests (265 values, 237 contacts)
+     * We need an interface to search for multiple values with "OR"
```
enkore: This was an intentional design decision to make scraping harder and to make it easier for ops to implement efficient rate limiting. The rationale is that this is pretty much a one-off action and therefore ok if it takes a few moments.

enkore:
I'm wondering, since this is implemented on top of qabel.core.index, that it maybe should be a separate package to minimize risk of mixing the layers?

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
```diff
@@ -0,0 +1,226 @@
...
+class MainIndexInteractor(private val indexServer: IndexServer,
+                          private val contactRepository: ContactRepository,
+                          private val identityRepository: IdentityRepository) : IndexInteractor {
+
+    /**
+     * TODO How to handle verifications?
+     * 1. save as Not verified -> Save update action? (simple way : add phoneVerified and emailVerified with Enum to identity)
+     * 2. trigger updateVerifyFlags on accept verification by received code -> Add inputCode, resendVerification Dialog on AppStartUp
+     * 3. Poll for verification updates while local data is not verified, work with verification timeouts?
+     */
```
enkore: The whole verification process is completely opaque to the requester (this piece of code). This was made so to avoid leaking activity of the user whose data is submitted to the requester, which doesn't have to be the user, by making it indistinguishable whether or even when a user has denied a (bogus) request.

The only way to find whether a request succeeded is to check whether the request was carried out, ie. the data in the public index was updated. Since the plan for q-android seems to be that the SMS code is entered in the app this is a good spot to poll it, esp. since in that case you know if (at least one code) was confirmed/denied. If one is denied then the entire request is forfeit. If the request also had an email then the app can't really know (unless the user chooses to copy the code from the email for some reason instead of opening the link in the email).

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
enkore: --- Re. OR instead of AND for multiple parameters (as it is currently implemented). In principle the reasoning of scraping/rate limiting applies also here. Or we make multiple parameters OR, but require that each parameter be of a different type. That would mean only 1 request per contact (phone=XXX OR mail=yyy).

enkore: The decision regarding scraping/rate-limiting could also be reversed, since it was recently contemplated to require authentication of index access. But that has not been decided yet. So maybe stick with this for now and we'll see about changes later?

core/src/main/java/de/qabel/core/index/errors.kt
```diff
@@ -45,7 +46,7 @@ internal fun errorMessageFromResponse(response: HttpResponse): String {
         JsonParser().parse(json)["error"].string
     } catch (e: Exception) {
         var msg = "No error message available (%s)".format(e)
-        if (e is JsonSyntaxException) {
+        if (e is JsonSyntaxException || e is NoSuchElementException) {
```
enkore: Good catch!

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
JoGir: Jop, authentications sounds good. Current version works for now...

chat/src/main/java/de/qabel/chat/repository/sqlite/adapter/ChatDropMessageAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+import de.qabel.core.repository.sqlite.hydrator.BaseEntityResultAdapter
+import java.sql.ResultSet
+
+class ChatDropMessageAdapter() : BaseEntityResultAdapter<ChatDropMessage>(ChatDropMessageDB) {
+
+    private val shareAdapter = ChatShareAdapter()
+
+    override fun hydrateEntity(entityId: Int, resultSet: ResultSet, entityManager: EntityManager, detached: Boolean): ChatDropMessage {
+        val payloadType = ChatDropMessageDB.toEnum(ChatDropMessage.MessageType.values(),
+            resultSet.getString(ChatDropMessageDB.PAYLOAD_TYPE.alias())!!, { it.type })
```
audax: throw a better exception than a null pointer

chat/src/main/java/de/qabel/chat/repository/sqlite/adapter/ChatShareAdapter.kt
```diff
@@ -0,0 +1,42 @@
...
+        with(resultSet) {
+            val status = getInt(STATUS.alias())
+            val hash = getString(ChatShareDB.HASH_TYPE.alias())?.let {
+                Hash(Hex.decode(getString(ChatShareDB.HASH.alias())), it)
+            }
+            val key = getString(ChatShareDB.KEY.alias())?.let {
+                SymmetricKey.Factory.fromHex(it)
+            }
+            return BoxFileChatShare(
+                ShareStatus.values().find { it.type == status }!!,
```
audax: throw a better exception than a null pointer.

audax:
Why does a PR about index-integration care so much about chat-messages and repositories?

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
```diff
@@ -0,0 +1,249 @@
...
+                        warn("Received unexpected UpdateResult[${it.name}] for delete action!")
+                        throw IndexServerException("Failed to delete old field value!")
+                    }
+                }
+            }
+        }
+        val currentStatus = findStateForValue(identity, newValue, fieldType)
+        return when (currentStatus) {
+            VerificationStatus.NOT_VERIFIED ->
+                UpdateIdentity(identity, listOf(UpdateField(UpdateAction.CREATE, fieldType, newValue!!))).let {
```
audax: there was never a not-null check for newValue…

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
```diff
@@ -0,0 +1,249 @@
...
+                updated = true
+            }
+
+            //Set nick from external contact
+            if (localContact.nickName.isNullOrBlank() || localContact.alias == localContact.nickName) {
+                localContact.nickName = receivedContact.nickName
+                updated = true
+            }
+
+            //TODO Force update for alias, email, phone?
```
audax: todos should be in a ticket or at least a comment in the PR

core/src/main/java/de/qabel/core/index/MainIndexInteractor.kt
JoGir: NOT_VERIFIED -> NotNull and NotBlank

enkore:
https://github.com/Qabel/qabel-index/issues/29

JoGir:
created ticket for qualified exceptions #597

JoGir:
> why does a PR about index-integration care so much about chat-messages and repositories?
Update IdentityRepository to BaseRepository and improved base classes.
IdentityRepository allow to get detached instances of Identities to check for changes.

core/src/test/java/de/qabel/core/ui/EntityUIExtensionTest.kt
```diff
@@ -0,0 +1,73 @@
...
+                val spaces = it.count { it.isWhitespace() }
+                assertThat(spaces, equalTo(4))
+            }
+        }
+    }
+
+    @Test
+    fun testReadableKeyShort(){
+        val readableKey = contact.readableKeyShort()
+        println(readableKey)
```
audax: 2 println in the test

core/src/test/java/de/qabel/core/ui/EntityUIExtensionTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    fun testReadableKey(){
+        val readableKey = contact.readableKey()
+        println(readableKey)
+        val lines = readableKey.lines()
+        assertThat(lines, hasSize(5))
+        lines.forEach {
+            if(lines.last() != it){
```
audax: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/drop-last.html

lines.dropLast(1).forEach

core/src/main/java/de/qabel/core/index/errors.kt
audax: fixed

JoGir:
TODO identity delete not deletes contact assignments

PR #594, opened 2016-09-13
Test for upgrading from old versions

audax:
The published release doesn't set a user_version, so any existing DMs will be migrated which crashes because the tables already exist.

See https://github.com/Qabel/qabel-android/issues/646


audax:
So. That happened.

julianseeger:
Nice workaround but do you have an idea how this actually happened and if this may happen on released versions?

audax:
@julianseeger exactly that. We had a schema before we used PRAGMA user_version, so when you converted the DM-stuff to the AbstractClientDatabase it always tried to apply the first migration because user_version=0 for old databases.

We didn't catch that error earlier because we also had the rootRef compatability error.

Murphys law.

PR #605, opened 2016-09-22
fix reupdate field in indexService

JoGir:


julianseeger:
don't think we need that code anymore (because create = update in the future) ... we only need a delete if new value = nullOrBlank

PR #607, opened 2016-09-24
Fix sharing

### LINKED ISSUE #608: FileMetadata broken on android
Android require an other version adapter

JoGir:
Resolves issues for android integration
fixes #608

### DIFF
--- a/box/src/main/java/de/qabel/box/storage/jdbc/JdbcFileMetadataFactory.kt
+++ b/box/src/main/java/de/qabel/box/storage/jdbc/JdbcFileMetadataFactory.kt
@@ -14,11 +14,11 @@ import java.sql.Connection
 import java.sql.DriverManager
 import java.sql.SQLException
 
-class JdbcFileMetadataFactory(val tmpDir: File) : FileMetadataFactory {
+class JdbcFileMetadataFactory @JvmOverloads constructor(
+    val tmpDir: File,
+    var versionAdapterFactory : (connection : Connection) -> VersionAdapter = { PragmaVersionAdapter(it)}
+) : FileMetadataFactory {
 
-    companion object {
-        var versionAdapterFactory : (connection : Connection) -> VersionAdapter = { PragmaVersionAdapter(it)}
-    }
 
     @Throws(QblStorageException::class)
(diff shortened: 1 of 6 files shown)


box/src/main/java/de/qabel/box/storage/jdbc/JdbcFileMetadataFactory.kt
```diff
@@ -5,12 +5,21 @@ import de.qabel.box.storage.BoxFile
...
 import java.io.File
 import java.io.IOException
+import java.sql.Connection
 import java.sql.DriverManager
 import java.sql.SQLException
 
 class JdbcFileMetadataFactory(val tmpDir: File) : FileMetadataFactory {
+
+    companion object {
+        var versionAdapterFactory : (connection : Connection) -> VersionAdapter = { PragmaVersionAdapter(it)}
```
julianseeger: Can't this be a constructor argument?

audax: Could be done in the same way as with the DirectoryMetadata

JoGir: and the sharingService requires a VersionAdapter then? i confirm this isnt a nice way, but this needs some more refactoring then change this to the constructor...

audax: The SharingService then requires a FileMetadataFactory which on android must be configured to use the android version adapter

audax: the factory can default to the factory from the core and with @JVMOverloads the default constructor wouldn't change. On Android you can pass the factory in from dagger.

julianseeger: What's wrong with constructor injection?

audax: It now works with constructor injection

Issue #496, opened 2016-04-20
cross client sync 


jmt85:
The Qabel core includes the main logic for the cross client sync. Every update of the a contact is a separate transaction. 

The core must also have the ability to solve conflicts (example via Rollbacks)

julianseeger:
Do we need to split the transaction files into separate directories? (For example per contact? Otherwise we would get a lot of changeset-files in a single directory)

julianseeger:
New approach for "special" files and folders:
we create a new prefix with a new indexDM and add all "special" stuff to that volume.

This way we don't need any special treatment for these files and they never appear in a filebrowser or folder sync.

audax:
If we use `SHA256(prefix + private_key +1)` for the adress of it, we won't get conflicts with box-index dms.

PR #610, opened 2016-09-26
#496 commands publish their boxObject

julianseeger:
- [x] wrapped AbstractNavigation::changes into SerializedSubject for thread safety
 - [x] CreateFolderChange publishes its' BoxObject (required in desktop)
 - [x] added a test to check that dm changes are pushed (on execute and not only on refresh)
 - [x] fixed `refresh` / `detectDmChanges` to notify about contents of new directories with a visitor

### DIFF
--- a/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
+++ b/box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
@@ -16,6 +16,7 @@ import org.apache.commons.codec.binary.Hex
 import org.apache.commons.lang3.NotImplementedException
 import org.spongycastle.crypto.params.KeyParameter
 import rx.lang.kotlin.PublishSubject
+import rx.subjects.SerializedSubject
 import rx.subjects.Subject
 import java.io.*
 import java.security.InvalidKeyException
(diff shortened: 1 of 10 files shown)


box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -185,34 +200,41 @@ abstract class AbstractNavigation(
         // remote folder adds
         newDm.listFolders()
             .filter { !originalDm.hasFolder(it.name) }
-            .map { CreateFolderChange(this, it.name, folderNavigationFactory, directoryFactory) }
+            .map { newFolders.add(it); it }
```
audax: That is not a map, that is a foreach

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
```diff
@@ -185,34 +200,41 @@ abstract class AbstractNavigation(
...
 
         // remote file changes (update, neither add nor delete)
         newDm.listFiles()
             .filter { originalDm.hasFile(it.name) && !hashEquals(originalDm.getFile(it.name)!!, it) }
-            .map { UpdateFileChange(originalDm.getFile(it.name)!!, it) }
+            .map { fileChange(it) }
             .forEach { push(it) }
     }
 
+    private fun fileChange(file: BoxFile) = UpdateFileChange(originalDm.getFile(file.name)!!, file)
```
audax: throw a better exception than a nullpointer if something bad happens ( ?: throw IllegalState...)

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
julianseeger: replaced by `.loop`

box/src/main/java/de/qabel/box/storage/AbstractNavigation.kt
julianseeger: ✓

PR #611, opened 2016-09-26
Wizard364

taynaps:
identityRepository observable


build.gradle
```diff
@@ -44,6 +44,8 @@ subprojects {
         testCompile 'org.mockito:mockito-core:2.0.52-beta'
         testCompile 'com.nhaarman:mockito-kotlin:0.6.0'
         compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+        compile group: 'com.googlecode.libphonenumber', name: 'libphonenumber', version: '3.+'
```
audax: Why is this needed? I thought the phone number lib is already available in the core.

audax: fixed

PR #621, opened 2016-09-28
 don't delete identities contacts associations on update

julianseeger:
fixes Qabel/qabel-desktop#487

### DIFF
--- a/core/src/test/java/de/qabel/core/repository/inmemory/InMemoryIdentityRepository.kt
+++ b/core/src/test/java/de/qabel/core/repository/inmemory/InMemoryIdentityRepository.kt
@@ -18,11 +18,11 @@ open class InMemoryIdentityRepository : IdentityRepository, EntityObservable by 
     private val identities = Identities()
 
     @Throws(EntityNotFoundException::class)
-    override fun find(id: String): Identity {
-        if (identities.getByKeyIdentifier(id) == null) {
-            throw EntityNotFoundException("id $id not found")
+    override fun find(keyId: String): Identity {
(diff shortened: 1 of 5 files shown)


core/src/test/java/de/qabel/core/repository/inmemory/InMemoryIdentityRepository.kt
```diff
@@ -18,11 +18,11 @@ open class InMemoryIdentityRepository : IdentityRepository, EntityObservable by
     private val identities = Identities()
 
     @Throws(EntityNotFoundException::class)
-    override fun find(id: String): Identity {
-        if (identities.getByKeyIdentifier(id) == null) {
-            throw EntityNotFoundException("id $id not found")
+    override fun find(keyId: String): Identity {
```
julianseeger: ftr.: this change was suggested by the IDE because the parameter `keyId` from the overwritten method was named `id` in the implementation which may cause problems when calling by named parameter. Unprobably but errorprone anyways

audax:
Everthing that doesn't have an explicit test, is broken…always.

Fix looks good and very reasonable.

We should check the usages of the "dangerous method" and see that nobody calls it like this without exactly the intention of removing all assignements.

julianseeger:
>We should check the usages of the "dangerous method" and see that nobody calls it like this without exactly the intention of removing all assignements.

We could split it into two methods: removeAssignments(contact) and updateAssignments(contact, List<Identity>) ... or only allow single calls like `removeAssignment(contact, identity)` and `addAssignment(contact, identity)` .... but at least in this case it was irritating but not THAT dramatic / dangerous ;)

PR #631, opened 2016-10-17
metaVolumes for #496 

julianseeger:
includes the new box prefix choosing algorithm:
 * prefer main prefix (prefix known by the local identity, that is assign to the current account)
 * use first prefix that exists locally and remote and has an index / root
 * use first prefix that exists remote and has an index / root
 * use first prefix that exists locally and remote (and has no index / root => create index)
 * create new prefix (and index)

Prefixes are chosen separately for the types USER and CLIENT


box/src/main/java/de/qabel/box/storage/factory/BlockBoxVolumeFactory.kt
```diff
@@ -0,0 +1,52 @@
...
+import de.qabel.core.accounting.BoxClient
+import de.qabel.core.config.Account
+import de.qabel.core.config.Identity
+import de.qabel.core.config.Prefix
+import de.qabel.core.repository.IdentityRepository
+import java.io.File
+import java.io.IOException
+import java.net.URI
+import java.net.URISyntaxException
+import java.nio.file.Files
```
julianseeger: gnah

audax:
Make this a default parameter, on android we have to inject the temp directories

box/src/main/java/de/qabel/box/storage/factory/AbstractBoxVolumeFactory.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    protected abstract fun writeBackend(prefix: String): StorageWriteBackend
+    protected abstract fun readBackend(prefix: String): StorageReadBackend
+
+    @JvmOverloads
+    fun choosePrefix(identity: Identity, account: Account, type: Prefix.TYPE = USER): String {
+        return PrefixChooser(
+            identityRepository,
+            { createNewPrefix(identity, boxClient, identityRepository) },
+            { prefix -> hasIndex(identity, prefix, type) },
```
julianseeger: addressed

PR #642, opened 2016-10-30
casted, filtered event streams

julianseeger:
to avoid reoccuring code like
```JAVA
dispatcher
    .events()
    .filter(e -> e is MessageReceivedEvent)
    .map(e -> (MessageReceivedEvent)e)
    .nowTheCodeICareAbout()
```
that can now be written as
```JAVA
dispatcher
    .events(MessageReceivedEvent.class)
    .nowTheCodeICareAbout()
```

### DIFF
--- a/core/src/main/java/de/qabel/core/event/EventDispatcher.kt
+++ b/core/src/main/java/de/qabel/core/event/EventDispatcher.kt
@@ -5,7 +5,7 @@ import rx.Observable
 interface EventSource {
     fun events(): Observable<Event>
     fun <T : Event> events(type: Class<T>): Observable<T>
-        = events().filter { type.isInstance(it) }.map { type.cast(it) }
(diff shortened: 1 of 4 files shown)


core/src/main/java/de/qabel/core/event/EventDispatcher.kt
```diff
@@ -4,6 +4,8 @@ import rx.Observable
 
 interface EventSource {
     fun events(): Observable<Event>
+    fun <T : Event> events(type: Class<T>): Observable<T>
+        = events().filter { type.isInstance(it) }.map { type.cast(it) }
```
audax: you can use the builtin ofType operator: http://reactivex.io/RxJava/javadoc/rx/Observable.html#ofType(java.lang.Class)

julianseeger: fixd

PR #654, opened 2016-11-01
Qabel/qabel-desktop#474 account carries its token

julianseeger:


audax:
Where does it get persisted?

julianseeger:
client specific as the passwort does currently. can you point the android code?

audax:
That is in code that wasn't touched for quite a while. In the android app, the token is just saved in the application private settings on the first login and accessed directly in the http backend classes:
https://github.com/Qabel/qabel-android/blob/master/qabelbox/src/main/java/de/qabel/qabelbox/communication/BaseServer.java#L97

audax:
I also somehow thought that this PR was in the desktop repo.

julianseeger:
Similar for the desktop client. Someone sets the account to the config (currently, the LoginController does config.setAccount(...) after the credentials have been entered) and the config(interface) is implemented persistent https://github.com/Qabel/qabel-desktop/blob/master/client/src/main/java/de/qabel/desktop/config/RepositoryBasedClientConfig.java#L67

So for the first step, the token would get persisted side-by-side with the password which is removed in another issue that needs to make the whole login dialog optional.

Issue #615, opened 2016-09-27
rework MainIndexService assumptions


julianseeger:
The MainIndexService makes some assumptions that don't make sense from the clients perspective, those should be checked and fixed.

For example:
Removing the identity is only possible with an email or phonenumber that is neither null nor blank. So I can't delete a previously uploaded identity when I previously deleted those fields ... but a user that no longer wants to be found would probably do exactly that and thus, could never delete his identity using the MainIndexService.
The scenarios get much worse if the identity gets deleted and re-imported...

enkore:
Yeah, it would be easier if we just allowed to say "remove *everything* you have on this identity". (This is server side.)

enkore:
Scope:

- Review MIS assumptions
- Align MIS model with server model
- Depends qabel/qabel-index#43 and qabel/qabel-index#44

thechauffeur:
Let us talk about this next time.

PR #660, opened 2016-11-14
Update index server

### LINKED ISSUE #600: Update index server API
(A specification update will also be prepared)

Gist:

- Search semantics changed

  - Instead of ANDing the terms, they are ORed.
  - Many terms can be specified (`email=...&email=....&phone=...`)
  - JSON POST is now allowed:

           {"query": [
              {"field": "email", "value": "abc@example.net"},
              ...
            ]}

  - JSON POST is recommended
  - Existing code should continue to work
  - Results now contain what matched each identity:

          {"identities": [
               {"public_key": ..........., "matches": [
                    {"field": "email", "value": "abc@example.net"},
                    ...
               ]}
          ]}

- Client authorization will be required on official servers

  - Exact same mechanism as the block server uses.

- Phone numbers are now normalized. This shouldn't change anything for well-behaved clients that already normalize everything to E.164.

enkore:
- [x] Fix IndexHTTPTest
- [x] Document how to update identity properties (drop URL and alias), also add a test for that
- [x] Update search endpoint to use the JSON API of the server
- [x] Add search API to search for a bunch of contacts at once; update IndexContact to have the "matches" property to indicate which contact in the result set belongs to which part of the query
- [x] Add API to query identity status
- [x] Add API to delete identity entirely
- [ ] Reimplement MainIndexService (in progress...)
- [x] Deduplicate retry logic
- [x] Nullable IdentityStatus
- [ ] Fix MainIndexService test
- [ ] Perhaps create lower level tests for the new APIs (<-- most of the existing ones are actually for validating correct use and understanding of gson)

Fixes #600 and implements the new APIs discussed in #615 meeting.

### DIFF
--- a/core/src/main/java/de/qabel/core/index/models.kt
+++ b/core/src/main/java/de/qabel/core/index/models.kt
@@ -138,7 +138,7 @@ data class IdentityStatus(
     /**
      * Current identity data as returned by server (drop URL, alias, public key).
      */
-    val identity: IndexContact,
+    val identity: IndexContact?,
     /**
      * A list of field statuses for every confirmed or unconfirmed (pending) entry associated with the
      * identity.
--- a/core/src/main/java/de/qabel/core/index/models.kt
+++ b/core/src/main/java/de/qabel/core/index/models.kt
@@ -19,7 +19,8 @@ import java.util.*
 data class IndexContact(
     val publicKey: QblECPublicKey,
     val dropUrl: DropURL,
-    val alias: String
+    val alias: String,
+    val matches: List<Field>  = listOf()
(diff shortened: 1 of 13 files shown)


enkore:
@JoGir: Can you have a look at the IndexHTTP API?

JoGir:
* Nullable IdentityStatus if not registered
* duplicate retry handling
* apply changes to IndexService

audax:
I think the index server in the test-docker-container is not updated @rehwanne

rehwanne:
I rebuilt the Container with new Versions of everything.

PR #667, opened 2017-01-11
client module

julianseeger:
makes sense, doesn't it?


enkore:
(Does this belong to #646?)

